PR #13, opened 2022-08-16
Fix minimal install and tidy up

ericpre:
### Progress of the PR
- [x] Fix minimum install,
- [x] Allow to run test suite without hyperspy and add corresponding build to CI - most of the tests are skipped but this is useful to test import without hyperspy being installed,
- [x] Remove deprecated `EMD` class in emd plugin,
- [x] Fix teardown `test_empad.py` module,
- [x] remove leftover `traits` dependency, 
- [x] add tests,
- [x] ready for review.


rsciio/tests/test_emd_prismatic.py
```diff
@@ -20,9 +20,11 @@
 
 import numpy as np
 import pytest
+
+hs = pytest.importorskip("hyperspy.api", reason="hyperspy not installed")
+
 import traits.api as t
```
jlaehne, line 26: remaining traits dependency (`rsciio` does not depend on `traits`, but I guess it is covered if `hs` is imported)

jlaehne (commented):
Great that the coverage could also be increased notably.

ericpre (commented):
I will leave the coverage improvement for another PR, as this one is already large enough and this is not the primary concern of this PR.

PR #23, opened 2022-08-22
Code improvements/consistency

jlaehne:
### Description of the change
Code improvements identified using [lgtm.org](https://lgtm.com/projects/g/hyperspy/rosettasciio?mode=tree&severity=&id=)

For "Modification of parameter with default" see explanation at https://lgtm.com/rules/4840097/

Otherwise, commit messages should be self explanatory. Mainly removes unused lines of code (unused imports, variables, or unreachable lines of code).

A few errors in the code turned up along the way.

### Progress of the PR
- [n/a] add tests,
- [x] ready for review.

### DIFF
--- a/rsciio/usid_hdf5/api.py
+++ b/rsciio/usid_hdf5/api.py
@@ -430,11 +430,12 @@ def file_reader(
         if not isinstance(dataset_path, str):
             raise TypeError("'dataset_path' should be a string")
         h5_dset = h5_f[dataset_path]
-        if not lazy:
-            h5_f.close()
-        return _usidataset_to_signal(
+        signal = _usidataset_to_signal(
             h5_dset, ignore_non_uniform_dims=ignore_non_uniform_dims, lazy=lazy, **kwds
         )
+        if not lazy:
+            h5_f.close()
+        return signal
 
 
 def file_writer(filename, object2save, **kwds):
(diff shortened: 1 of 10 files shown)


rsciio/usid_hdf5/api.py
```diff
@@ -432,10 +432,6 @@ def file_reader(
             h5_dset, ignore_non_uniform_dims=ignore_non_uniform_dims, lazy=lazy, **kwds
         )
 
-    # At least close the file handle if not lazy load
-    if not lazy:
```
ericpre, line 436: When not loading lazily, we still need to close the file.

ericpre:
I am surprised that these errors doesn't appear in https://github.com/hyperspy/rosettasciio/security/code-scanning?

rsciio/usid_hdf5/api.py
jlaehne: But these two lines are unreachable code:
https://lgtm.com/projects/g/hyperspy/rosettasciio?mode=list&id=py%2Funreachable-statement

The preceding `if/else` are both ending with a `return` statement, so the file closing would not be executed. So I guess, we have to close the file before the return statement.

jlaehne: Implemented.

jlaehne:
> I am surprised that these errors doesn't appear in https://github.com/hyperspy/rosettasciio/security/code-scanning?

I am indeed startled as well, because both should be using the same codeQL engine to scan the code.

rsciio/usid_hdf5/api.py
jlaehne: ok, done

ericpre: To simplify the code, would it be better to keep only one return at the end of the function?

PR #25, opened 2022-08-23
Add reader for Jobin-yvon .xml-format

pietsjoh:
### Description of the change
Add support for reading the .xml-format from jobin-yvon.

### Progress of the PR
- [x] basic load functionality (data, axes, original_metadata, metadata) + corresponding tests 
- [x] load signal data
    - [x] check order of map (which direction is X and which is Y)
- [x] implement different signal axis formats (with corresponding units)
    - [x] wavelength (nm)
    - [x] wavenumber (1/cm)
    - [x] abs. wavenumber (1/cm)
    - [x] energy (eV)
    - [x] different names for abs. and rel. wavenumber?
- [x] metadata mapping
    - [x] remove spaces in nodes
    - [x] add more entrys from original_metadata
    - [x] rotation angle
    - [x] glued spectrum
- [x] add tests (currently dependant on hyperspy.load())
- [x] ready for review.

```python
from rsciio.jobin_yvon import api
api.file_reader("test.xml")
```


rsciio/jobin_yvon/api.py
```diff
@@ -0,0 +1,779 @@
...
+            self.metadata["Acquisition_instrument"]["Spectrometer"][
+                "pinhole"
+            ] = self.metadata["Acquisition_instrument"]["Spectrometer"][
+                "entrance_slit_width"
+            ]
+        except KeyError:
+            pass
+
+
+def file_reader(file_path, use_uniform_wavelength_axis=True, **kwds):
```
jlaehne, line 753: Most readers use `filename` as parameter.

rsciio/jobin_yvon/api.py
```diff
@@ -0,0 +1,779 @@
...
+    Each element can have the following attributes: attrib, tag, text.
+    Moreover, non-leaf-elements are iterable (iterate over child-nodes).
+    In this specific format, the tags not useful to extract information.
+    Instead, the "ID"-entry in attrib is used to identify the sort of information.
+    The IDs are consistent for the tested files.
+
+    Parameters
+    ----------
+    file_path: pathlib.Path
+        path to the to be read file
```
jlaehne, line 42:
```suggestion
        Path to the to be read file.
```
The common style is to use full sentences here. (see https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard )

rsciio/jobin_yvon/api.py
```diff
@@ -0,0 +1,779 @@
...
+
+    use_uniform_wavelength_axis: bool
```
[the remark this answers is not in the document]
jlaehne: I would use `use_uniform_signal_axis` to be more generic.
Mention if parameter is `optional`, give default value in the description of the parameter.

rsciio/jobin_yvon/api.py
```diff
@@ -0,0 +1,779 @@
...
+            return "image"
+        else:
+            return ""
+
+    @property
+    def _signal_type(self):
+        if self._lumispy_installed:
+            return "Luminescence"
+        else:
+            return ""
```
jlaehne, line 492: I would use the generic `Signal1D` as backup `signal_type`

rsciio/jobin_yvon/specifications.yaml
```diff
@@ -0,0 +1,7 @@
+format_name: 'Jobin Yvon'
+description: 'Reads data from Jobin Yvon xml files'
+full_support: False
+file_extensions: ['xml', 'XML']
+default_extension: 0
+writes: False
+non_uniform_axis: False
```
jlaehne, line 7:
```suggestion
non_uniform_axis: False

```

rsciio/jobin_yvon/specifications.yaml
```diff
@@ -0,0 +1,7 @@
+format_name: 'Jobin Yvon'
```
jlaehne, line 1: I would consider calling it 'JobinYvon' without space.

rsciio/jobin_yvon/specifications.yaml
```diff
@@ -0,0 +1,7 @@
+format_name: 'JobinYvon'
+description: 'Reads data from Jobin Yvon xml files'
```
jlaehne, line 2:
```suggestion
description: 'Read data from .xml files saved using Horiba Jobin Yvon's LabSpec software.'
```

upcoming_changes/25.new.rst
```diff
@@ -0,0 +1 @@
+Add support for reading the ``.xml``-format from Jobin-Yvon.
```
jlaehne, line 1:
```suggestion
Add support for reading the ``.xml``-format from Horiba Jobin Yvon's LabSpec software.
```

jlaehne:
@LMSC-NTappy as you have implemented the `.sur` reader not so long ago, may you have a look at this PR?

docs/supported_formats/jobin_yvon.rst
```diff
@@ -0,0 +1,56 @@
...
+also uses the ``.xml`` file-extension.
+
+The reader supports all signal axis units exported by LabSpec, i.e. wavelengths,
+wavenumbers (absolute), Raman shift (relative wavenumbers),
+as well as energy. 
+
+.. Note::
+
+  The wavelength-to-energy conversion is not documented for LabSpec, i.e. it is
+  not clear whether the permittivity of air is taken into account. When working
```
LMSC-NTappy, line 33: From what I see, the term "permittivity of air" is largely used in the codebase, e.g.:

https://github.com/LumiSpy/lumispy/blob/4669de6846c30ab40b65ee0308b1b7812b96f5b3/lumispy/utils/axes.py#L34

This is fine but as far as I can see, correction is applied using the real-valued refractive index. Just something to be aware of...

rsciio/jobin_yvon/api.py
```diff
@@ -0,0 +1,819 @@
...
+            pass  # pragma: no cover
+
+        ## add percentage for filter key name
+        new_filter_key_name = "ND Filter (%)"
+        try:
+            self.original_metadata["experimental_setup"][
+                new_filter_key_name
+            ] = self.original_metadata["experimental_setup"]["ND Filter"]
+            del self.original_metadata["experimental_setup"]["ND Filter"]
+        except KeyError:
```
LMSC-NTappy, line 274: To me, this seems like an overly complicated way of writing

```python
self.original_metadata["experimental_setup"]["ND Filter (%)"] = \ 
    self.original_metadata["experimental_setup"].pop("ND Filter")
```

If you consider adding a default to the field such as NA, the try / except block could be removed altogether.

rsciio/jobin_yvon/api.py
```diff
@@ -0,0 +1,819 @@
...
+        (use_uniform_wavelength_axis).
+
+        Parameters
+        ----------
+        array: np.ndarray
+            Contains the axes data points.
+        name: Str
+            Name of the axis.
+        """
+        scale = np.abs(array[0] - array[-1]) / (array.size - 1)
```
LMSC-NTappy, line 347: I like the approach of loading as uniform axis and warn the user of possible non-uniformity. 

However, I would use a regression method such as numpy.polyfit in order to minimize the linearized axis error:
`[scale, offset] = np.polyfit(np.arange(array.size),array,1)`

rsciio/jobin_yvon/api.py
```diff
@@ -0,0 +1,819 @@
...
+            if rel_diff_compare > 0.01 and self._use_uniform_signal_axis:
+                _logger.warning(
+                    f"The relative variation of the {name}-axis-scale ({rel_diff_compare:.3f}%) exceeds 1%.\n"
+                    "                              "
+                    f"Difference between first 2 entrys: {abs_diff_begin:.3f}.\n"
+                    "                              "
+                    f"Difference between last 2 entrys: {abs_diff_end:.3f}.\n"
+                    "                              "
+                    "Using a non-uniform-axis is recommended."
+                )
```
LMSC-NTappy, line 361: Here, is there any specific reason to compare the first / last two entries? 

I would intuitively warn if the max error exceeds the threshold.
```python
rel_diff_compare  = (np.diff(array) - scale)/scale
if rel_diff_compare > 0.01 and self._use_uniform_signal_axis:
    ...
```

rsciio/jobin_yvon/api.py
```diff
@@ -0,0 +1,819 @@
...
+            self.metadata["Signal"] = {}
+        if "Sample" not in self.metadata:
+            self.metadata["Sample"] = {}
+        if "Acquisition_instrument" not in self.metadata:
+            self.metadata["Acquisition_instrument"] = {
+                "Laser": {"Filter": {}, "Polarizer": {}},
+                "Spectrometer": {"Grating": {}, "Polarizer": {}},
+                "Detector": {"processing": {}},
+                "Spectral_image": {},
+            }
```
LMSC-NTappy, line 526: Here, are all the if statement really needed? They are bound to evaluate to True since the metadata object is newly created and thus empty, if I am right.

rsciio/jobin_yvon/api.py
```diff
@@ -0,0 +1,819 @@
...
+
+        self.metadata["Signal"]["signal_type"] = self._signal_type
+        self.metadata["Signal"]["signal_dimension"] = 1
+
+        self._set_metadata(
+            self.metadata["Sample"],
+            "description",
+            self.original_metadata["file_information"],
+            "Sample",
+        )
```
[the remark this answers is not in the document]
LMSC-NTappy: I agree. 

Mostly, I dislike the `_set_metadata` paradigm, because in essence all it does is assigning in a rather obscure way:
`self.metadata["Sample"]["description"] = self.original_metadata["file_information"]["Sample"]`

I would rather do the assignment explicitly with None values if keys don't exist, and then remove the None.

e.g.
```python
Laser = {}
Laser['wavelength'] = self.original_metadata["experimental_setup"].get("Laser (nm)")
Laser["objective_magnification"] = self.original_metadata["experimental_setup"].get("Objective")
#etc etc
self.metadata["Acquisition_instrument"]["Laser"] = {k: v for k, v in Laser.items() if v is not None}
```

rsciio/jobin_yvon/api.py
```diff
@@ -0,0 +1,819 @@
...
+            pass  # pragma: no cover
+
+
+def file_reader(filename, use_uniform_signal_axis=False, **kwds):
+    """Read data from .xml files saved using Horiba Jobin Yvon's LabSpec software.
+
+    Parameters
+    ----------
+    use_uniform_signal_axis: bool, default=True
+        Can be specified to choose between non-uniform or uniform signal-axis.
```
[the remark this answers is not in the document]
LMSC-NTappy: Agreed, except it is not the average delta that is used (see my commant above)

LMSC-NTappy (changes requested):
Overall looks good to me.

I have some troubles with the way the non uniform axes are linearized, which can lead to errors in some cases.

Other than that, I suggested a few changes that in my opinion improve readability

docs/supported_formats/jobin_yvon.rst
jlaehne: Indeed, we should be consistent and use `refractive index of air`. For the other occurences I have opened #147.

LMSC-NTappy: Nice. Thank you for this,

pietsjoh: I changed that for jobin_yvon.rst.

rsciio/jobin_yvon/api.py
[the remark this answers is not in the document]
pietsjoh: I like the idea of using .get() and removing None values. This is now implemented.

rsciio/jobin_yvon/api.py
pietsjoh: Originally, I didn't create a new metadata object in this method. Hence the if statements. But now they are not necessary anymore and I deleted them.

rsciio/jobin_yvon/api.py
pietsjoh: I simplified this code block as you suggested. But I don't want to use default values for the dict. As this would complicate the metadata mapping.

rsciio/jobin_yvon/api.py
pietsjoh: In the test datasets (array[i+1] - array[i]) decreases from the first 2 to the last 2 values. But checking for the max error makes generally more sense. So I will change that.

rsciio/jobin_yvon/api.py
pietsjoh: An advantage of the method I used is that the boundaries match exactly.
Moreover, the offset obtained from the fit is wrong when the axis contains negative values.
For example there is a test file with navigation axis (-2, 0, 2).

I personally wouldn't use the fit for the offset calculation, but I could imagine using a fit for the scale.

rsciio/jobin_yvon/api.py
jlaehne: > Also, a 1% threshold certainly seems very large to me but feel free to let it that way if you think that's the way to go

In the beginning I thought that as well, but even with that threshold it would warn on most of our test datasets. That is why the idea is to default to a non-uniform axis, but allow the loading as a uniform axis as an option.

rsciio/jobin_yvon/api.py
jlaehne: Indeed, the only additional degree of freedom of the fit should be that the first and last point are not fixed. So if we want to keep their positions, there is no advantage of the fit.

rsciio/jobin_yvon/api.py
LMSC-NTappy: Ok :)

rsciio/jobin_yvon/api.py
LMSC-NTappy: > An advantage of the method I used is that the boundaries match exactly.

> Indeed, the only additional degree of freedom of the fit should be that the first and last point are not fixed. So if we want to keep their positions, there is no advantage of the fit.

I don't see any virtue in ensuring that first/last points match exactly but if that's a concern to you I guess you have good reasons.

> Moreover, the offset obtained from the fit is wrong when the axis contains negative values.
For example there is a test file with navigation axis (-2, 0, 2).

Sounds odd to me. Running this gives
```python
np.polyfit(x=np.arange(3),y=[-2,0,2],deg=1)
#out: array([ 2., -2.])
```
Which is exactly a scale of 2 and an offset of -2

> I personally wouldn't use the fit for the offset calculation, but I could imagine using a fit for the scale.

Definitely this would induce more error than what is currently implemented (at lease in some cases)

You do as you please :)

rsciio/jobin_yvon/api.py
LMSC-NTappy: Ok :)

Defaulting to a non-uniform axis is an even better option I think. 

I am not sure how general is the scope of the Jobin Yvon XML file format. If it's dedicated to spectroscopy then cubic axes (1/cubic for the energy-like representation) would cover everything.

LMSC-NTappy (approved):
Looks good to me.

Linearisation strategy is still a subject of disagreement but if by default the signal is loaded as nua I guess it is the user's responsibility to linearize in the way it pleases him/her (and decide how the linearization error is dealt with).

Cheers

Nico

rsciio/jobin_yvon/api.py
pietsjoh: > Which is exactly a scale of 2 and an offset of -2

I made a mistake here. Yes it should also work in this case.

I use a fit now to get the scale/offset of the signal axis.

LMSC-NTappy (approved):
Looks good to me!

IMHO there is nothing preventing this to be merged.

PR #28, opened 2022-08-24
activate additional queries

jlaehne:
### Description of the change
The CodeQL scan did not find the errors and warnings currently listed by lgtm.org as noticed in https://github.com/hyperspy/rosettasciio/pull/23#issuecomment-1226322856
Therefore, added [additional queries over the standard ones](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#using-queries-in-ql-packs)

### Progress of the PR
- [x] ready for review.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -51,8 +51,7 @@ jobs:
         # Prefix the list here with "+" to use these queries and those in the config file.
         
         # Details on CodeQL's query packs refer to : https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#using-queries-in-ql-packs
-        # queries: security-extended,security-and-quality
-        queries: security-extended
+        queries: +security-extended,security-and-quality
 
         
     # Autobuild attempts to build any compiled languages  (C/C++, C#, or Java).
--- a/rsciio/_hierarchical.py
+++ b/rsciio/_hierarchical.py
@@ -785,7 +785,7 @@ class HierarchicalWriter:
             else:
                 try:
                     group.attrs[key] = value
-                except BaseException:
+                except Exception:
                     _logger.exception(
                         "The writer could not write the following "
                         f"information in the file: {key} : {value}"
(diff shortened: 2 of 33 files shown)


jlaehne:
Now it has a lot of additional alerts: https://github.com/hyperspy/rosettasciio/security/code-scanning?query=is%3Aopen+pr%3A28

We might evaluate whether we want both `security-extended` and `security-and-quality` or just the former.

jlaehne:
Merging #23 into this one already closed 318 of the alerts

jlaehne:
Using `BaseException` instead of the sub-class `Exception` may lead to `KeyboardInterrupt` being missed: https://github.com/hyperspy/rosettasciio/security/code-scanning/330
Should we change these occurences as proposed? What is your oppinion @ericpre?

jlaehne:
All other remaining alerts need to be reviewed individually, as some might point to an existing issue, but many will be false alerts or unrelevant.

ericpre:
> Using `BaseException` instead of the sub-class `Exception` may lead to `KeyboardInterrupt` being missed: [Except block handles 'BaseException'](/hyperspy/rosettasciio/security/code-scanning/330) Should we change these occurences as proposed? What is your oppinion @ericpre?

As far as I understand, in most cases, we should use `Exception` to let `KeyboardInterrupt` being raised and use `BaseException` only when we *do want* to catch `KeyboardInterrupt` and do something about it, so they need to be changed.

jlaehne:
> As far as I understand, in most cases, we should use `Exception` to let `KeyboardInterrupt` being raised and use `BaseException` only when we _do want_ to catch `KeyboardInterrupt` and do something about it, so they need to be changed.

Added a commit, I did not find any point where I would have left it as `BaseException`.

jlaehne:
I would propose to merge this PR to have code alert comments for new PRs available. Remaining alerts from the existing code can be addressed in a follow-up PR.

ericpre (approved):
After merging #59, it now needs a rebase!

PR #30, opened 2022-08-29
Remove deprecation Bruker, EMD, Nexus

### LINKED ISSUE #24: Remove deprecated code
A number of readers have deprecation warnings, usually for specific keywords -- sometimes HyperSpy v2.0 is explicitly mentioned as the point where they will be removed. We could consider to remove these deprecated keywords already before releasing the first version of RosettaSciIO as it would be a good occasion to do so and the library should be first used by HyperSpy 2.0 anyway.

jlaehne:
### Description of the change
Remove deprecated keywords and attributes from readers before first release (which will be the one used by HyperSpy 2.0, for which these deprecations were announced).

Closes #24

### Progress of the PR
- [X] Change implemented (can be split into several points),
- [X] update user guide (if appropriate),
- [X] add an changelog entry in the `upcoming_changes` folder (see [`upcoming_changes/README.rst`](https://github.com/hyperspy/rosettasciio/blob/main/upcoming_changes/README.rst)),
- [X] remove tests,
- [X] ready for review.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,20 +8,6 @@ https://rosettasciio.readthedocs.io/en/latest/changes.html
 
 .. towncrier release notes start
 
-API Removal
------------
-
-Following the deprecation cycle announced in `HyperSpy
-<https://hyperspy.org/hyperspy-doc/current/user_guide/changes.html>`_,
-the following keywords and attributes have been removed:
-
-- :ref:`Bruker composite file (BCF) <bcf-format>`: The 'spectrum' option for the
-  `select_type` parameter was removed. Use 'spectrum_image' instead.
-- :ref:`Electron Microscopy Dataset (EMD) NCEM <emd_ncem-format>`: Using the
-  keyword 'dataset_name' was removed, use 'dataset_path' instead.
-- :ref:`NeXus data format <nexus-format>`: The `dataset_keys`, `dataset_paths`
-  and `metadata_keys` keywords were removed. Use `dataset_key`, `dataset_path`
-  and `metadata_key` instead.
 
 Initiation (2022-07-23)
 =======================
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,8 +11,9 @@ https://rosettasciio.readthedocs.io/en/latest/changes.html
 API Removal
 -----------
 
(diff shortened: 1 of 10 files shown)


CHANGES.rst
```diff
@@ -8,6 +8,20 @@ https://rosettasciio.readthedocs.io/en/latest/changes.html
 
 .. towncrier release notes start
 
+API Removal
+-----------
```
ericpre, line 12: As this is relevant to hyperpsy users and the API was in hyperspy before, would it be better to move it to the corresponding entry of hyperpsy 2.0 changelog?

rsciio/tests/test_nexus_hdf.py
```diff
@@ -26,9 +26,8 @@
 import traits.api as t
 import h5py
 
-from hyperspy.exceptions import VisibleDeprecationWarning as HSVisibleDeprecationWarning
+from hyperspy.exceptions import VisibleDeprecationWarning as VisibleDeprecationWarning
```
ericpre, line 29:
```suggestion
from hyperspy.exceptions import VisibleDeprecationWarning
```

CHANGES.rst
jlaehne: Good point, we might want to just duplicate it there, in case migrating HyperSpy users end up looking here?

jlaehne: see https://github.com/hyperspy/hyperspy/pull/3002

ericpre: I still find odd to have a `API removal` section for a first release! I would remove this section and a normal changelog entry - the description on its own is good.

rsciio/tests/test_nexus_hdf.py
ericpre: What's about this?

jlaehne: I commited it online, but then overwrote it when pushing the updated version from my PC ;-) Should be fixed now.

CHANGES.rst
jlaehne: I moved it to a regular towncrier entry for the PR and checked the formatting locally.

ericpre (approved):
Thanks!

Issue #37, opened 2022-09-10
Signal type of EDS dataset


ericpre:
We should improve the situation with setting the `signal_type` of EDS dataset, make it consistent across EDS data file reader and document its behaviour in `rosettasciio` and in `hyperspy`.

The approach used in the `bruker` reader seems to be the most sensible:
https://github.com/hyperspy/rosettasciio/blob/57fccc4bcc19945c7a33d1d4351ee57260f9305f/rsciio/bruker/api.py#L1496-L1511

In any case, data acquired on a SEM doesn't guaranty that the `signal_type` should be `EDS_SEM`, as the specimen can be thin enough to make `EDS_TEM` more suitable!

sem-geologist:
SEM_EDS and TEM_EDS should go away from `rosettasciio`, why? 1) because this is feature creap to io library, 2) brakes compatibility with i.e. emsa formats (format saved with hyperspy is not compatible with DTSA-II or NeXLSpectrum.jl), 3) complicates (rather blocks) the life for other methods which wanted for long time already to extend on this library: i.e. μXRF, but I can see other methods like PIXE and Xray- inducing posibilities are comming - it is insane to have separate signal type, while it is Xrays gathered by same type of detector. SEM, TEM, XRF, PIXE - those are interpretations of datangathered by spectrometer which does not care. Make this non-mandatorry mode as atribute - not the root of metadata structure!

jlaehne:
To my understanding, the `metadata->Signal->signal_type` is actually an attribute and not the root of the metadata structure.

Users of HyperSpy, of course still will want that the `signal_type` is set correctly when reading the files. For the moment, I don't see a good way to keep that out of RosettaSciIO completely, otherwise the HyperSpy-IO functions will need again a format-specific code that decides on what signal-type to choose (and for the moment all the format-specific stuff has been moved to this repository).

> brakes compatibility with i.e. emsa formats

If a `file_writer` does not adhere to format definitions, we should of course correct that. EMSA is a clearly defined format specification, where we should not add anything that breaks compatibility. If so, you should open a separate issue on that.

ericpre:
I suspect that there isn't a perfect solution to this problem (not setting a signal type at all is not convenient for a vast majority of users) and we need to be pragmatic with what is the most appropriate approach. I opened the issue so that we can discuss about it! :)

[...]

PR #39, opened 2022-09-16
Fix MSA signaltype

jlaehne:
### Description of the change
As brought up in #37, the current implementation of the MSA writer does not comply to the format definitions according to
https://the-mas.org/wp-content/uploads/2018/11/emmff_ascii.txt

In particular, HyperSpy `signal_types` where written to MSA 1:1, even if they deviate from the allowed strings according to the format definition.

This PR correctly converts HyperSpy signal strings to the corresponding MSA ones when writing. On the reading side, the resulting `signal_types` don't necessarily correspond to the HyperSpy ones. Correct conversion to signal types of HypersSpy has to be done on the HyperSpy/LumiSpy side (e.g. choosing which voltage corresponds to SEM or TEM) as proposed in #37.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs
s = hs.load("rosettasciio/rsciio/tests/msa_files/minimum_metadata.msa")
s.metadata.Signal.signal_type = "EDS_SEM"
s.save("tmp.msa")
s2 = hs.load("tmp.msa")
```
Note that this example can be useful to update the user guide.

### DIFF
--- a/rsciio/msa/api.py
+++ b/rsciio/msa/api.py
@@ -287,7 +287,7 @@ def parse_msa_string(string, filename=None):
             mapped.set_item("Signal.signal_type", "EELS")
         elif parameters["SIGNALTYPE"] == "CLS":
             mapped.set_item("Signal.signal_type", "CL")
-        else:
+        else:  # pragma: no cover
             if parameters["SIGNALTYPE"] not in [
                 "EDS",
                 "WDS",
--- a/rsciio/msa/api.py
+++ b/rsciio/msa/api.py
@@ -20,6 +20,7 @@ from datetime import datetime as dt
 import codecs
 import os
 import logging
+import warnings
 from turtle import dot
 
 import numpy as np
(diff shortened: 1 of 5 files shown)


ericpre (approved):
Should it have a changelog entry?

rsciio/msa/api.py
```diff
@@ -281,14 +282,28 @@ def parse_msa_string(string, filename=None):
...
-        if mapped.Signal.signal_type == "ELS":
-            mapped.Signal.signal_type = "EELS"
-        if mapped.Signal.signal_type in ["EDX", "XEDS"]:
-            mapped.Signal.signal_type = "EDS"
+    if "SIGNALTYPE" in parameters and parameters["SIGNALTYPE"]:
+        if parameters["SIGNALTYPE"] == "ELS":
+            mapped.set_item("Signal.signal_type", "EELS")
+        elif parameters["SIGNALTYPE"] == "CLS":
+            mapped.set_item("Signal.signal_type", "CL")
+        else:
```
ericpre, line 290:
```suggestion
        else:  # pragma: no cover
```

PR #55, opened 2022-10-27
Add reader for renishaw .wdf format

pietsjoh:
This reader is based on py-wdf-reader with extra insights from gwyddion.
Compared to py-wdf-reader, more metadata is extracted (PSET Blocks).
Moreover, changes are made to meet rosettasciio's format.

### Progress of the PR
- [x] Adapt to general rosettasciio structure
- [x] Extend metadata extraction
- [x] add tests
- [x] map metadata
- [x] look at WHTL, WMAP parsing
- [x] update docstring (if appropriate)
- [x] update user guide (if appropriate)
- [x] add an changelog entry in the `upcoming_changes` folder (see [`upcoming_changes/README.rst`](https://github.com/hyperspy/rosettasciio/blob/main/upcoming_changes/README.rst))
- [x] ready for review

```python
from rsciio.msa import api
api.file_reader("file.wdf")
```


sem-geologist:
@pietsjoh, what is still missing for this to be ready for review?

sem-geologist:
I see in code there is marked some lines with caveats that there could be unknown and unreadable structures. Is this based on Reverse engineering? I think this format is potentially moving target. In such case it would be easier to write parsing using kaitai_struct as it is much easier to adapt parsing then when binary file specification gets updated by OEM.

I have experience with RE such moving target (i.e. https://github.com/sem-geologist/peaksight-binary-parser). I think it would be then useful to communicate with gwyddion, and have single Reverse engineering implementation written in kaitai, where gwyddion would compile it into C++ library and this would use python compilation of code. BTW there are some free kaitai_struct descriptions for jpeg and EXIF parsing definitions ready, which could be reused then without needing to add PIL as optional dependency. 

I saw there is PIL checked as optional dependency to extract jpeg and EXIF; What about using `imagio` instead as that is one of requirements for this library?

pietsjoh:
> I see in code there is marked some lines with caveats that there could be unknown and unreadable structures. Is this based on Reverse engineering?

Yes, this format is reverse engineered. I based my work on the [py-wdf-reader](https://github.com/alchem0x2A/py-wdf-reader) with some extra inspiration from [gwyddion](https://github.com/christian-sahlmann/gwyddion/blob/master/modules/file/renishaw.c).

>I think this format is potentially moving target. In such case it would be easier to write parsing using kaitai_struct as it is much easier to adapt parsing then when binary file specification gets updated by OEM.

>I have experience with RE such moving target (i.e. https://github.com/sem-geologist/peaksight-binary-parser). I think it would be then useful to communicate with gwyddion, and have single Reverse engineering implementation written in kaitai, where gwyddion would compile it into C++ library and this would use python compilation of code.

I have no experience with kaitai yet, but this might be an option in the future. However, I don't know if Renishaw changes their format often enough for this to be useful.

> I saw there is PIL checked as optional dependency to extract jpeg and EXIF; What about using imagio instead as that is one of requirements for this library?

My implementation is close to how py-wdf-reader reads the EXIF-Tags. I tried using imageio instead, but couldn't get it to work. The picture with the exif-data is still included in the metadata (even when PIL is not installed). So I figured that it is not a big problem if I leave it like this. But I can give it another go and try to use imageio for reading the EXIF tags.

> @pietsjoh, what is still missing for this to be ready for review?

The main issue is licensing, as I based my work on the py-wdf-reader. I just asked in https://github.com/alchem0x2A/py-wdf-reader/issues/43 how to handle this. Moreover, I asked Renishaw directly for some insights on some of the TODO's/questions I have. But this could also be implemented in a different PR.

jlaehne:
> > @pietsjoh, what is still missing for this to be ready for review?
> 
> The main issue is licensing, as I based my work on the py-wdf-reader. I just asked in [alchem0x2A/py-wdf-reader#43](https://github.com/alchem0x2A/py-wdf-reader/issues/43) how to handle this. Moreover, I asked Renishaw directly for some insights on some of the TODO's/questions I have. But this could also be implemented in a different PR.

I would consider this ready to review ... the remaining questions can be handled along the way :-)

docs/supported_formats/renishaw.rst
```diff
@@ -0,0 +1,16 @@
...
+
+Renishaw
+--------
+
+Reader for spectroscopy data saved using Renishaw's WiRE software.
+Currently, RosettaSciIO can only read the ``.wdf`` format from Renishaw.
+
+If `LumiSpy <https://lumispy.org>`_ is installed, ``Luminescence`` will be
+used as the ``signal_type``.
+
```
jlaehne, line 11: Mention that development is based on the `py-wdf-reader`, the matlab package and the Gwyddion module also here and not only in the header of the file.

Add note that tested on a limited set of data files.

pietsjoh: I extended the documentation. Do you have any extra suggestions?

pietsjoh:
I tried loading the exif tags with `imageio` and remember now why I stayed with `PIL`. `imageio` returns exif as a binary sequence, whereas `PIL` returns exif as a dict (tag number as key and tag as value). I can easily deal with the dict. However, I would need to write an extra parser for the binary sequence.

docs/supported_formats/renishaw.rst
```diff
@@ -0,0 +1,28 @@
...
+   There are many different options for the axes according to the format specifications.
+   However, only a limited subset is tested: `Spectral` (Wavelength and Raman Shift) for
+   the signal axes and `X`, `Y`, `Z`, `FocusTrackZ` and `Time` for navigation axes.
+   Reading Maps obtained in a serpentine path is not implemented.
+
+
+This reader is based on the `py-wdf-reader <https://github.com/alchem0x2A/py-wdf-reader.git>`_,
+which is inspired by the `matlab reader <https://doi.org/10.5281/zenodo.495477>`_ from Alex Henderson.
+Moreover, inspiration is took from `gwyddion's reader <http://gwyddion.net>`_.
```
jlaehne, line 22:
```suggestion
Moreover, inspiration is taken from `gwyddion's reader <http://gwyddion.net>`_.
```

rsciio/renishaw/_api.py
```diff
@@ -0,0 +1,1332 @@
...
+    %s
+    %s
+    use_uniform_signal_axis: bool, default=False
+        Can be specified to choose between non-uniform or uniform signal axes.
+        If `True`, the ``scale`` attribute is calculated from the average delta
+        along the signal axis and a warning is raised in case the delta varies
+        by more than 1%%.
+    load_unmatched_metadata: bool, default=False
+        Some of the original_metadata cannot be matched (no key just value).
+        Part of this is a VB-Script used for data acquisition (~230kB),
```
jlaehne, line 1303:
```suggestion
        Part of this is a VisualBasic-Script used for data acquisition (~230kB),
```

rsciio/renishaw/_api.py
```diff
@@ -0,0 +1,1332 @@
...
+        which blows up the original_metadata size. If this option is set to
```
jlaehne, line 1304:
```suggestion
        which blows up the size of ``original_metadata``. If this option is set to
```

rsciio/renishaw/_api.py
```diff
@@ -0,0 +1,1332 @@
...
+        `True` this metadata will be included and can be accessed by
```
jlaehne, line 1305:
```suggestion
        `True`, this metadata will be included and can be accessed by
```

rsciio/renishaw/_api.py
```diff
@@ -0,0 +1,1332 @@
...
+        s.original_metadata.UNMATCHED, otherwise the UNMATCHED tag will not exist.
```
jlaehne, line 1306:
```suggestion
        ``s.original_metadata.UNMATCHED``, otherwise the ``UNMATCHED`` tag will not exist.
```

ericpre:
It needs a rebase and change the test file location: https://rosettasciio.readthedocs.io/en/latest/contributing.html#adding-and-updating-test-data 

> I have experience with RE such moving target (i.e. https://github.com/sem-geologist/peaksight-binary-parser). I think it would be then useful to communicate with gwyddion, and have single Reverse engineering implementation written in kaitai, where gwyddion would compile it into C++ library and this would use python compilation of code. BTW there are some free kaitai_struct descriptions for jpeg and EXIF parsing definitions ready, which could be reused then without needing to add PIL as optional dependency.

Thank you @sem-geologist for sharing this. This is indeed very interesting if it would enable us to share code across programming language. However, from trying to understand how easy it would be to use by reading the documentation, it seems that the python tooling for kaitai is still early stage and not mature! Maybe this is one of these, where it is worth putting effort in getting the tooling to a stage where it can be used easily, before starting to use it in rosettasciio?
Practically, it will take time to implement it here and in the meantime, we should go ahead with the current code of this PR.

ericpre (commented):
I didn't review the code thoroughly but it looks in good shape. The test file location need to be changed:
https://rosettasciio.readthedocs.io/en/latest/contributing.html#adding-and-updating-test-data

sem-geologist:
> Thank you @sem-geologist for sharing this. This is indeed very interesting if it would enable us to share code across programming language. However, from trying to understand how easy it would be to use by reading the documentation, it seems that the python tooling for kaitai is still early stage and not mature! 

Where had You get that from? Go, and Rust are not mature in kaitai_struct translations, but python... It is exceptionally well supported. It has not only official functional translator for parsing, but have an alternative translator using python `Construct` instead of python library `kaitaistruct`, which allows read/write (parsing and serializing back) functionality of binary structures. Version <1.0 does not mean that it is not mature.

pietsjoh:
pre-commit.ci autofix

jlaehne (approved):

PR #47, opened 2022-09-19
Doc and import improvement

ericpre:
As discussed in https://github.com/hyperspy/rosettasciio/issues/38. This changes the import and docstring/documentation for the `bruker` and `blockfile` format as examples for discussion.

### Progress of the PR
- [ ] Change implemented (can be split into several points),
- [ ] update docstring (if appropriate),
- [ ] update user guide (if appropriate),
- [ ] add an changelog entry in the `upcoming_changes` folder (see [`upcoming_changes/README.rst`](https://github.com/hyperspy/rosettasciio/blob/main/upcoming_changes/README.rst)),
- [ ] Check formatting changelog entry in the `readthedocs` doc build of this PR (link in github checks)
- [ ] add tests,
- [ ] ready for review.

### Minimal example of the bug fix or the new feature
```python
from rsciio.msa import api
api.file_reader("your_msa_file.msa")
# Your new feature...
```
Note that this example can be useful to update the user guide.

Issue #38, opened 2022-09-12
Consolidate docstring and documentation


ericpre:
The problem is also, that it is not consistent how detailed the `file_reader` and `file_writer` docstrings are. For some formats more is documented in the user guide, for others more in the docstrings.

_Originally posted by @jlaehne in https://github.com/hyperspy/rosettasciio/pull/35#discussion_r968012172_

ericpre:
One approach could be to:
- put the generic/introductory plugin specific documentation a `rst`/`md` file for each plugin, which could live in the `docs` folder or in the plugin folder
- add the `file_reader` and `file_write` docstrings in the plugin specific page of the docs

This could be done by one of the two:
- use a sphinx template
- include the API reference in the plugin specific documentation page.

Considering that sphinx is a bit brittle and not always clearly documented, using the sphinx template approach may be more difficult but it would have the advantage of enforcing consistence between plugins and avoid repeating code.

jlaehne:
>     * put the generic/introductory plugin specific documentation a `rst`/`md` file for each plugin, which could live in the `docs` folder or in the plugin folder

That file exists for every plugin afaik, in `docs/supported_formats/`.
 
>     * add the `file_reader` and `file_write` docstrings in the plugin specific page of the docs

That remains to be done, though some plugins have very detailed docstrings and others none at all.

PR #59, opened 2022-11-05
Doc and import improvement #2

jlaehne:
### Description of the change
Follow up on #47, adresses #38. Starting to continue working on the readers (from the back of the alphabet).

### Progress of the PR
- [ ] add tests,
- [ ] ready for review.

### DIFF
--- a/rsciio/tiff/_api.py
+++ b/rsciio/tiff/_api.py
@@ -29,6 +29,12 @@ import numpy as np
 from tifffile import imwrite, TiffFile, TIFF
 from tifffile import __version__ as tiffversion
 
+from rsciio.docstrings import (
+    FILENAME_DOC,
+    LAZY_DOC,
+    RETURNS_DOC,
+    SIGNAL_DOC,
+)
 from rsciio.utils.tools import DTBox, _UREG
 from rsciio.utils.date_time_tools import get_date_time_from_metadata
 
@@ -62,11 +68,8 @@ def file_writer(filename, signal, export_scale=True, extratags=[], **kwds):
 
     Parameters
     ----------
-    filename: str
-        Filename of the file to write to.
-    signal: dict
-        Dictionary containing the signal object as defined in the `API guide
-        <https://hyperspy.org/rosettasciio/api.html>`_.
+    %s
+    %s
     export_scale: bool, default=True
         Export the scale and the units (compatible with DM and ImageJ) to
         appropriate tags.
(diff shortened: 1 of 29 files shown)


jlaehne:
Working on this PR I realized that plugins have to be loaded by their `folder name`, which is often not the same as the `plugin name`:
USID - usid_hdf5
Semper - semper_unf
DigitalSurfSurface - sur

When loading in HyperSpy, the reader can be specified by the plugin name, but when loading an individual plugin one has to use the `folder_name`, which is not mentioned anywhere in the documentation so far! For example, `from rsciio import usid_hdf5`.

We may want to give the folders more consistent names that correspond to the default plugin name (capitalization is another question here) and then document the names in the table of supported formats.

rsciio/tiff/_api.py
```diff
@@ -112,16 +138,40 @@ def file_reader(filename, force_read_resolution=False, lazy=False, **kwds):
...
+          Initialise a DataAxis.
+        - ``functional``: the best-fit 3rd-order polynomial axis is used,
+          avoiding linearisation error. Initialise a FunctionalDataAxis.
+
+        By default, ``uniform`` is used but a warning of the linearisation error
+        is issued. Explicitly passing ``hamamatsu_streak_axis_type='uniform'``
+        suppresses the warning. In all cases, the original axis values are stored
+        in the ``original_metadata`` of the signal object.
+    **kwds: optional
+        Additional arguments to be passed to the `tifffile library
```
ericpre, line 163:
```suggestion
        Additional arguments to be passed to the ``TiffFile`` class of the `tifffile library
```

rsciio/tiff/_api.py
```diff
@@ -58,16 +64,33 @@
...
+    %s
+    export_scale: bool, default=True
         Export the scale and the units (compatible with DM and ImageJ) to
         appropriate tags.
+    extratags: tuple, list of tuples, optional
+        Save custom tags through the ``tifffile`` library. Must conform to a
+        specific convention (see `tifffile documentation
+        <https://github.com/cgohlke/tifffile/#examples>`_ and example below).
+    **kwds: optional
+        Additional arguments to be passed to the `tifffile library
```
ericpre, line 81:
```suggestion
        Additional arguments to be passed to the ``imwrite`` function of the `tifffile library
```

ericpre (approved):

PR #67, opened 2022-11-17
Bugfix : error when reading JEOL .pts file with frame list 

nem1234:
### Description of the change
- Bugfix : error when reading JEOL .pts file with frame list 
- Add test for frame list and frame_start_index
- Add test for frameshift

### Progress of the PR
- [x] Change implemented (can be split into several points),
- [ ] update docstring (if appropriate),
- [ ] update user guide (if appropriate),
- [ ] add an changelog entry in the `upcoming_changes` folder (see [`upcoming_changes/README.rst`](https://github.com/hyperspy/rosettasciio/blob/main/upcoming_changes/README.rst)),
- [ ] Check formatting changelog entry in the `readthedocs` doc build of this PR (link in github checks)
- [x] add tests,
- [ ] ready for review.

### Minimal example of the bug fix or the new feature
```python
import hyperspy.api as hs
s = hs.load("file.pts", frame_list=[1,3])
frame_start_index = s.original_metadata.jeol_pts_frame_start_index
s = hs.load("file.pts", frame_list=[5,2], frame_start_index=frame_start_index)
```

PR #68, opened 2022-11-22
Fix error when reading pts file with frame_list, frame_start_index (was PR #67)

nem1234:
### Description of the change

- Bugfix : error when reading JEOL .pts file with un-ordered frame list
- Bugfix : error when length of frame_start_index is smaller than sweep count
- Revise warning message
- Add test for frame list, frame_shift and frame_start_index
- Add test for broken asw file
- Add test for SEM data 

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```import hyperspy.api as hs
s = hs.load("file.pts", frame_list=[1,3])
frame_start_index = s.original_metadata.jeol_pts_frame_start_index
s = hs.load("file.pts", frame_list=[5,2], frame_start_index=frame_start_index)
```

### DIFF
--- a/rsciio/jeol/api.py
+++ b/rsciio/jeol/api.py
@@ -63,15 +63,13 @@ def file_reader(filename, **kwargs):
 
 def _read_asw(filename, **kwargs):
     image_list = []
-    fd = open(filename, "br")
-    file_magic = np.fromfile(fd, "<I", 1)[0]
-    if file_magic != 0:
-        _logger.warning("Not a valid JEOL asw format")
-        fd.close()
-        return []
-    fd.seek(12)
-    filetree = _parsejeol(fd)
-    fd.close()
+    with open(filename, "br") as fd:
+        file_magic = np.fromfile(fd, "<I", 1)[0]
+        if file_magic != 0:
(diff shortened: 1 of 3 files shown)


jlaehne:
Continuation of #67

ericpre (approved):
It looks good.

Do you want to add a release note mentioning the following?
> - Bugfix : error when reading JEOL .pts file with un-ordered frame list
> - Bugfix : error when length of frame_start_index is smaller than sweep count

nem1234:
Thank you.
This error seems very rare cases, so I don't think adding release note is needed.

By the way, if it is needed, should I write issue before adding upcoming_changes?
Is unnumbered log allowed?

ericpre:
Based on my experience, release notes are useful even for the most minor changes/bug fix, as it is only useful for users but also for ourselves in the future and developers of other libraries.

To add a release note, you just need to add a file in the `upcoming_changes` folder with the correct name (PR or issue number and type of changes) and then `towncrier` will process the entry automatically (add the release note in the right place and add create the hyperlink).

nem1234:
Ok,  I put 68.bugfix.rst into upcoming_change directory

PR #72, opened 2022-12-07
Doc and import improvement #4

jlaehne:
### Description of the change
Follow up on #49, #59, and #64 adresses #38. For readers with folder names from I...M+dens.

### Progress of the PR
- [ ] Change implemented (can be split into several points),
- [ ] update docstring (if appropriate),
- [ ] update user guide (if appropriate),
- [ ] add an changelog entry in the `upcoming_changes` folder (see [`upcoming_changes/README.rst`](https://github.com/hyperspy/rosettasciio/blob/main/upcoming_changes/README.rst)),
- [ ] Check formatting changelog entry in the `readthedocs` doc build of this PR (link in github checks)
- [ ] add tests,
- [ ] ready for review.

PR #64, opened 2022-11-15
Doc and import improvement #3

jlaehne:
### Description of the change
Follow up on #49 & #59, adresses #38. For readers with folder names from N...P.

### Progress of the PR
- [X] Change implemented (can be split into several points),
- [X] update docstring (if appropriate),
- [X] update user guide (if appropriate),
- [ ] add an changelog entry in the `upcoming_changes` folder (see [`upcoming_changes/README.rst`](https://github.com/hyperspy/rosettasciio/blob/main/upcoming_changes/README.rst)),
- [ ] Check formatting changelog entry in the `readthedocs` doc build of this PR (link in github checks)
- [X] add tests,
- [X] ready for review.

PR #76, opened 2023-01-16
Doc and import improvement #5

### LINKED ISSUE #38: Consolidate docstring and documentation

jlaehne:
### Description of the change
Follow up on #47, #59, #64, and #72. Closes #38. For readers with folder names from D...H. Final PR of this series.

### Progress of the PR
- [x] add tests,
- [x] ready for review.


jlaehne:
Since it is missing from the list of workflow runs, here is the link to the corresponding version of the documentation:
https://rosettasciio--76.org.readthedocs.build/en/76/

rsciio/edax/_api.py
```diff
@@ -123,7 +129,7 @@ def get_spc_dtype_list(load_all=False, endianess="<", version=0.61):
     """
     Get the data type list for an SPC spectrum.
     Further information about the file format is available `here
-    <https://github.com/hyperspy/hyperspy/files/29506/SPECTRUM-V70.pdf>`__.
+    <https://hyperspy.org/rosettasciio/_downloads/9e2f0ccf5287bb2d17f1b7550e1d626f/SPECTRUM-V70.pdf>`__.
```
ericpre, line 132: Do you know how "permanent" is this link, if it can change with new release, etc? Is it possible to use sphinx label as alternative?

docs/supported_formats/hspy.rst
```diff
@@ -3,43 +3,44 @@
...
 information will be lost in the writing process and that supports saving data
 of arbitrary dimensions. It is based on the `HDF5 open standard
 <https://www.hdfgroup.org/solutions/hdf5/>`_. The HDF5 file format is supported by `many
 applications
 <https://support.hdfgroup.org/products/hdf5_tools/SWSummarybyName.htm>`_.
-Part of the specification is documented in :external+hyperspy:ref:`metadata_structure`.
+Parts of the specifications are documented in :external+hyperspy:ref:`metadata_structure`.
 
-.. versionadded:: 1.2
+.. versionadded:: HyperSpy_v1.2
```
ericpre, line 15: Is there a reason for the underscore?
```suggestion
.. versionadded:: HyperSpy v1.2
```

docs/supported_formats/hspy.rst
```diff
@@ -3,43 +3,44 @@
...
     Enable saving HSpy files with the ``.hspy`` extension. Previously only the
     ``.hdf5`` extension was recognised.
 
-.. versionchanged:: 1.3
+.. versionchanged:: HyperSpy_v1.3
```
ericpre, line 19:
```suggestion
.. versionchanged:: HyperSpy v1.3
```

docs/supported_formats/hspy.rst
```diff
@@ -70,7 +71,12 @@ This feature is particularly useful when using
      <BaseSignal, title: X-ray line intensity of EDS SEM Signal1D: Mn_La at 0.63 keV, dimensions: (|)>,
      <BaseSignal, title: X-ray line intensity of EDS SEM Signal1D: Zr_La at 2.04 keV, dimensions: (|)>]
 
-.. versionadded:: 1.3.1
+.. _hspy-chunks:
+
+Chunking
+^^^^^^^^
+
+.. versionadded:: HyperSpy_v1.3.1
```
ericpre, line 79:
```suggestion
.. versionadded:: HyperSpy v1.3.1
```

rsciio/digital_micrograph/_api.py
```diff
@@ -1237,27 +1238,31 @@ def get_mapping(self):
...
-    contains more than one dataset.
+    If more than one dataset is contained in the ``.dm3/4`` file, a list of
+    signals is returned.
 
     Parameters
     ----------
-    record_by: Str
+    %s
+    %s
+    record_by: str
```
ericpre, line 1252: This may be something to follow up in a separate PR.

docs/supported_formats/edax.rst
```diff
@@ -10,43 +10,32 @@ spectrum image is loaded from the corresponding ``.ipr`` and ``.spc`` files
...
+
 When using `HyperSpy <https://hyperspy.org>`_, if elemental information has been
-defined in the spectrum image, those
-elements will automatically be added to the signal loaded by HyperSpy.
+defined in the spectrum image, those elements will automatically be added to the
+signal loaded by HyperSpy.
 
-In HyperSpy, if an ``.spd`` file is loaded, the result will be a Hyperspy
+In HyperSpy, if an ``.spd`` file is loaded, the result will be a HyperSpy
 ``EDSSEMSpectrum`` map, and the calibration will be loaded from the appropriate
```
ericpre, line 19: Note for later, before it gets lost: for a separate PR, it would be good to review references to HyperSpy API objects and link it correctly - it may need further improvement to the HyperSpy API, not sure if the attribute object are well documented at the moment in HyperSpy documentation.

ericpre (commented):
Thank you @jlaehne for finishing this up, I have left some suggestions.

docs/supported_formats/hspy.rst
jlaehne: Yes there is, only the first word after `.. versionadded::` is printed bold (or even printed in the header, I don't quite remember). Without `HyperSpy` it seems like we are referrinf to RosettaSciIO versions and without the underscore it is not clear, which HyperSpy version one refers to. Any alternative symbol would of course be fine.

rsciio/digital_micrograph/_api.py
jlaehne: Giving more details?

docs/supported_formats/edax.rst
jlaehne: Yes, we should create an issue for that. But the new documentation should maybe be released for that?

Also I found some missing labels when referencing from LumiSpy to HyperSpy, which currently require full links instead of proper references.

rsciio/digital_micrograph/_api.py
ericpre: Yes, maybe or whether, this is still used or not.

docs/supported_formats/edax.rst
ericpre: If this is the case, we should change the branch of the development version on readthedocs.org to use the `RELEASE_next_major` branch instead of the `RELEASE_next_minor` branch, as it make more sense to work from that at the moment. By doing this, we don't need to rely on a release version and we can tweak labels iteratively if necessary, etc.
I don't have the permission the readthedocs.org to change it, @francisco-dlp, can you please change this setting on readthedocs.org?

docs/supported_formats/hspy.rst
ericpre: ok, this sounds the best, then!

rsciio/edax/_api.py
jlaehne: I changed it to (the new) GitHub links again, which should be "permanent" when not moving the file.

Problem with sphinx label would be that it is readable only in the parsed docstring on read-the-docs and e.g. not in docstrings displayed inline in python.

rsciio/digital_micrograph/_api.py
jlaehne: It is saved into the metadata, but DM is the only reader that allows setting it as parameter. Normally, it should be determined from the metadata. So I will also remove it from here.

Values can be "spectrum" or "image".

rsciio/edax/_api.py
ericpre: > Problem with sphinx label would be that it is readable only in the parsed docstring on read-the-docs and e.g. not in docstrings displayed inline in python.

Indeed, but this is the same for all sphinx hyperlink? However, this is an issue with rendering and not the docstring itself - hopefully, https://github.com/jupyter/papyri will improve the situation, when it would ready to use!

rsciio/digital_micrograph/_api.py
ericpre: [`record_by` is deprecated since HyperSpy 1.2](https://hyperspy.readthedocs.io/en/latest/user_guide/metadata_structure.html#signal), maybe it should also be removed altogether from the metadata dictionnary returned by the reader?

rsciio/edax/_api.py
```diff
@@ -123,7 +129,7 @@ def get_spc_dtype_list(load_all=False, endianess="<", version=0.61):
     """
     Get the data type list for an SPC spectrum.
     Further information about the file format is available `here
-    <https://github.com/hyperspy/hyperspy/files/29506/SPECTRUM-V70.pdf>`__.
+    <https://github.com/hyperspy/rosettasciio/raw/main/docs/file_specification/edax/SPECTRUM-V70.pdf>`__.
```
ericpre, line 132:
```suggestion
    <https://github.com/hyperspy/rosettasciio/blob/57b114cc02a0cbe022f1c50c1d59779a44952685/docs/file_specification/edax/SPECTRUM-V70.pdf>`__.
```

rsciio/edax/_api.py
```diff
@@ -524,7 +530,7 @@ def get_ipr_dtype_list(endianess="<", version=333):
     """
     Get the data type list for an IPR image description file.
     Further information about the file format is available `here
-    <https://github.com/hyperspy/hyperspy/files/29507/ImageIPR.pdf>`__.
+    <https://github.com/hyperspy/rosettasciio/raw/main/docs/file_specification/edax/ImageIPR.pdf>`__.
```
ericpre, line 533:
```suggestion
    <https://github.com/hyperspy/rosettasciio/blob/57b114cc02a0cbe022f1c50c1d59779a44952685/docs/file_specification/edax/ImageIPR.pdf>`__.
```

rsciio/edax/_api.py
```diff
@@ -39,8 +45,7 @@ def get_spd_dtype_list(endianess="<"):
     """
     Get the data type list for an SPD map.
     Further information about the file format is available `here
-    <https://github.com/hyperspy/hyperspy/
-    files/29505/SpcMap-spd.file.format.pdf>`__.
+    <https://github.com/hyperspy/rosettasciio/raw/main/docs/file_specification/edax/SpcMap-spd.file.format.pdf>`__.
```
ericpre, line 48:
```suggestion
    <https://github.com/hyperspy/rosettasciio/blob/57b114cc02a0cbe022f1c50c1d59779a44952685/docs/file_specification/edax/SpcMap-spd.file.format.pdf>`__.
```

rsciio/edax/_api.py
ericpre: Actually, because this is for the docstring of `get_spc_dtype_list` and not `file_reader`, it doesn't appear in rosettasciio documentation, so it may be better to use with the permanent link from github.

rsciio/edax/_api.py
```diff
@@ -952,29 +950,50 @@ def spd_reader(
...
+        Otherwise, the name of the .ipr file to use for spatial calibration
+        can be explicitly given as a string.
     **kwargs
-        Additional keyword arguments supplied to the readers
-
-    Returns
-    -------
+        Remaining arguments are passed to the Numpy ``memmap`` function.
 
+    %s
```
ericpre, line 988: To give more visibility to available resources:

```suggestion
    %s

    Notes
    -----
    The file specification is available at :ref:`edax-file_specification`.
```

rsciio/digital_micrograph/_api.py
jlaehne: OK, then we should open a separate issue and PR, because there are still several readers setting this parameter!

jlaehne: see  #79

rsciio/edax/_api.py
jlaehne: > > Problem with sphinx label would be that it is readable only in the parsed docstring on read-the-docs and e.g. not in docstrings displayed inline in python.
> 
> Indeed, but this is the same for all sphinx hyperlink? However, this is an issue with rendering and not the docstring itself - hopefully, https://github.com/jupyter/papyri will improve the situation, when it would ready to use!

But I think, we don't have so many links in the docstrings. Mostly in the user-guide, where using the sphinx references makes sense.

rsciio/edax/_api.py
jlaehne: The `raw` link directly loads the file, the `blob` loads the github page of the file -- do you really prefer that one?

ericpre: The ["blob" link is permanent](https://docs.github.com/en/repositories/working-with-files/using-files/getting-permanent-links-to-files), and it should exist as long as the commit exists on github. The other link will break when its location changes in the repository, or the branch name changes, etc.

ericpre (approved):

PR #80, opened 2023-01-21
Change year in header to 2023

jlaehne:
### Description of the change
Change year in header to 2023.

### Progress of the PR
- [x] ready for review.

### DIFF
--- a/rsciio/__init__.py
+++ b/rsciio/__init__.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2007-2022 The HyperSpy developers
+# Copyright 2007-2023 The HyperSpy developers
 #
 # This file is part of RosettaSciIO.
 #
--- a/rsciio/_hierarchical.py
+++ b/rsciio/_hierarchical.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2007-2022 The HyperSpy developers
+# Copyright 2007-2023 The HyperSpy developers
 #
 # This file is part of RosettaSciIO.
 #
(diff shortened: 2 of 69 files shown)


ericpre:
Shall we finish https://github.com/hyperspy/rosettasciio/pull/76 and then update the renamed `edax/_api.py` here?

jlaehne:
Good point, it is actually 6 renamed `_api.py` files in #76.

Lets merge #76 first and then sort it out here.

jlaehne:
Updated after merge of #76

ericpre (commented):
Is it missing the change in `digital_micrograph/_api.py`?

jlaehne:
> Is it missing the change in `digital_micrograph/_api.py`?

Yes, this one has a more complicated copyright line than the others that is missed by the `grep` command. The initial work from Stefano Mazzucco is honored, but then the initial year of development within HyperSpy is different.

For many readers the development will probably have started later, but it will be hard to differentiate that. So maybe should adapt the line for the DM plugin as well to use the same line as in the other files, but add a mention of the initiation by Stefano Mazzucco?

ericpre:
I don't know, but the formatting of reader seems to be valid. Do you think that it's possible to find a pattern with grep which works for all?

jlaehne:
So far I did a `sed` replace for `Copyright 2007-2022` on lines containing `Copyright`, but I think just replacing `-2022` should suffice. I took a note and will see next year ;-)

hakonanes:
Have you tried `licenseheaders`, https://pypi.org/project/licenseheaders/? We use this in the pyxem packages. It might catch this more complicated copyright line.

Issue #79, opened 2023-01-21
Deprecate `record_by` metadata setting


jlaehne:
As discussed in https://github.com/hyperspy/rosettasciio/pull/76#discussion_r1083278185 the `record_by` parameter in the metadata is [deprecated since HyperSpy v1.2](https://hyperspy.readthedocs.io/en/latest/user_guide/metadata_structure.html#signal). However, several readers are still setting that parameter. Thus, we should make a PR to remove it, where remaining.

PR #102, opened 2023-04-19
Deprecate record_by metadata attribute

pietsjoh:
### Description of the change
This PR solves #79, removing the deprecated `record_by` metadata attribute for the affected file readers.

`record_by` was used in these cases to infer the `signal_dimension`, as the `navigate` attribute wasn't set for the axes.
Now `navigate` is set for all readers so that `record_by` can be removed.

### Progress of the PR
- [ ] add an changelog entry in the `upcoming_changes` folder
- [ ] ready for review.


jlaehne:
Could you merge in `main` @pietsjoh to see whether #106 fixed the currently failing tests?

upcoming_changes/102.deprecation.rst
```diff
@@ -0,0 +1 @@
+Remove record_by attribute from file readers where remaining
```
jlaehne, line 1:
```suggestion
Remove deprecated `record_by` attribute from file readers where remaining
```

ericpre (commented):
This looks good.
Can please you add tests to check that the signal and navigation are correct for the following reader:
- bruker
- digitalmicrograph
- ripple
- tiff

There are no tests for mrc... but that should be sorted in a separate PR.

ericpre:
@pietsjoh, do you want me to help with the rebase? The conflicts should come from https://github.com/hyperspy/rosettasciio/pull/123 and the change to the path of the test files, etc..

pietsjoh:
@ericpre yes, I am currently working on it

pietsjoh:
I missread your comment. I want to try it by myself first and I will ask you when I cannot make it work

ericpre:
Indeed, this hasn't done everywhere because it is not possible to use the same pattern, see for example: https://github.com/hyperspy/rosettasciio/pull/123#discussion_r1209080592 and https://github.com/hyperspy/rosettasciio/pull/123#discussion_r1207184411.

There were already a lot of changes in that PR, so I have left this part as it was.

pietsjoh:
ok, thanks for the clarification

Issue #101, opened 2023-04-19
failing tests after merging #100


jlaehne:
#### Describe the bug
Tests are failing after merging ~~#90~~ #100:
https://github.com/hyperspy/rosettasciio/actions/runs/4742423201/jobs/8420679897

ericpre:
Actually, this is after merging #100:
https://github.com/hyperspy/rosettasciio/actions/runs/4742402405

```
FAILED rsciio/tests/test_pantarhei.py::test_save_load_cycle - box.exceptions.BoxKeyError: "'<class 'rsciio.utils.tools.DTBox'>' object has no attribute scan_generator"
FAILED rsciio/tests/test_tiff.py::test_JEOL_SightX - TypeError: 'NoneType' object is not iterable
```

<details>
<summary>Full log</summary>

```
=================================== FAILURES ===================================
_____________________________ test_save_load_cycle _____________________________
[gw0] darwin -- Python 3.8.16 /Users/runner/hostedtoolcache/Python/3.8.16/x64/bin/python

tmp_path = PosixPath('/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pytest-of-runner/pytest-0/popen-gw0/test_save_load_cycle3')

    def test_save_load_cycle(tmp_path):
        fname = tmp_path / "test_file.prz"
    
        s = hs.load(my_path / "pantarhei_data" / "panta_rhei_sample_v5.prz")
>       s.save(fname)

rsciio/tests/test_pantarhei.py:85: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../hostedtoolcache/Python/3.8.16/x64/lib/python3.8/site-packages/hyperspy/signal.py:3043: in save
    io_save(filename, self, overwrite=overwrite, file_format=file_format, **kwds)
../../../hostedtoolcache/Python/3.8.16/x64/lib/python3.8/site-packages/hyperspy/io.py:921: in save
    importlib.import_module(writer["api"]).file_writer(
rsciio/pantarhei/_api.py:69: in file_writer
    data, meta_data = export_pr(signal=signal)
rsciio/pantarhei/_api.py:276: in export_pr
    meta_data = _metadata_converter_out(metadata, original_metadata)
rsciio/pantarhei/_api.py:373: in _metadata_converter_out
[...]

PR #107, opened 2023-04-25
Bugfix for  Issue #101 (failing tests after merging #100)

nem1234:
### Description of the change
- xml name space in JEOL SightX tiff is rewritten with '_'.

### Progress of the PR
- [ ] add tests,
- [x] ready for review.

```pytest tests/test_tiff
```

### DIFF
--- a/rsciio/tiff/_api.py
+++ b/rsciio/tiff/_api.py
@@ -540,15 +540,17 @@ def _is_jeol_sightx(op) -> bool:
 
 def _axes_jeol_sightx(tiff, op, shape, names):
     # convert xml text to dictionary of tiff op['ImageDescription']
-    # convert_xml_to_dict need to remove white spaces before decoding XML
-    scales, offsets, units = _axes_defaults()
+    import xml.etree.ElementTree as ET
+    from rsciio.utils.tools import XmlToDict
+    from box import Box
 
+    scales, offsets, units = _axes_defaults()
     jeol_xml = "".join(
         [line.strip(" \r\n\t\x01\x00") for line in op["ImageDescription"].split("\n")]
     )
-    from rsciio.utils.tools import convert_xml_to_dict
-
-    jeol_dict = convert_xml_to_dict(jeol_xml)
+    jeol_xml_obj = ET.fromstring(jeol_xml)
+    xml_to_dict = XmlToDict()
+    jeol_dict = Box(xml_to_dict.dictionarize(jeol_xml_obj))
     op["ImageDescription"] = jeol_dict["TemReporter"]
     eos = op["ImageDescription"]["Eos"]["EosMode"]
     illumi = op["ImageDescription"]["IlluminationSystem"]
--- a/upcoming_changes/101.bugfix.rst
+++ b/upcoming_changes/101.bugfix.rst
@@ -1,0 +1,1 @@
+For more robust xml to dict conversion, convert_xml_to_dict is replaced by XmlToDict (introduced by PR #111).


nem1234:
I'm not sure this change should be applied here, or convert_xml_to_dict in utils/tools.py.

jlaehne (approved):

sem-geologist:
using dots in namespace is standard XML practice, rather utils/tools.py needs fixing.

sem-geologist:
... but well... with all hones, namespaces are overused in latest JEOL *******ware and built-in python's ElementTree historically has reappearing problems with that (lxml handles those much better). While we were looking for new SEM to buy, we looked to Jeol offer. AFAIK and could depict from presented situation, Jeol moved its software stack from very clean design made in C to some "web on PC" - monstrosity thus embracing XML and verbose nonsense all over data files. That raised me a red flag that it will be thousands times to hell and back to implement any reading functionality of their new formats. Luckily they had not won the public procurement for other reasons. It dreads me to think about any implementation of that. I really salute You for trying to implement that and am worried that constant workaround will be needed to add due their mess... well...

ericpre:
Thanks @nem1234 and @sem-geologist. I wasn't sure what is the best way to approach this. This issue arises with `python-box` 7.0 which added support for creating nested dictionary from flat key, where the dots are used as separator.
I was concerned that just cleaning the string by replace the problematic key may not be future proof enough. @sem-geologist, based on your comment, it may be necessary to add an argument to `python-box` to disable the new behaviour with "flat keys" and dot separator.

nem1234:
I'm tring to rewrite tools.py, but I'm confusing how to build the dict tree .....

In this case, 
XML text:   "<A>x<B>b1</B><B>b2</B>y<C>c<D>d</D></C>z</A>"

Current code:
dictree = {'A': 'x'}
Only when XML structure is really simple, this is useful.

Case 1. Direct conversion from XML
dictree = {
  "A": [
    "x",
    { "B": ["b1", "b2"] },
    "y",
    {"C" : [ "c", { "D" : "d" } } ],
    "z"
    ] }

dictree["A"] **[1]** ["B"][0] == "b1"

cost to access data is almost the same as direct access with XML library.
need a extra code to find the nested key "B"
Using XML library is better than this.

Case 2.  
{"A": { "_text": "xyz", "B": [ {"_text":"b1"}, {"_text","b2" ], "C":{"_text":"c", "D": {"_text":"d"} } } }

dictree["A"]["B"][0]["_text"] == "b1"

Text values mixed with elements are merged.
This seems better than case 1, but need ["_text"] to access the values.

Case 3.
{"A" { "B": ["b1","b2"], "C":{"D":"d"}}}
dictree["A"]{"B"][0] == "b1"

Text values mixed with elements are ignored.
I think this is enough.

nem1234:
When decoding JEOL tiff file, list structure is needed. (Currently it is overwritten with the last value.
`<A><B>b1</B><B>b2</B></A>`
dictree["A"]["B"] == "b2"  ("b1") is overwritten.

sem-geologist:
> cost to access data is almost the same as direct access with XML library

Well that sums the main disadvantage of dictionary-zation of XML.

If we would need to deal with single XML files or few, then yes this whole endeavor would make of little sense and better would be to make wrapper tool for unified access to XML or dictionaries. (or use some ready tools to do that, i.e. `lxml`s `objectify` looks nice method)

However... to load XML the Whole DAMN verbose XML file (or IO stream) needs to be read into memory. Some of OEM just loves (I would with whole confidence wold say they pathologically enjoy it) extremely verbose XML which more than 50% contains absolutely useless information. And then we want to concatenate information from many such XML it is very easy to run out of memory. Thus benefit of pythonization of XML's is that we copy relevant data discarding all the junk. The Cost of accessing a single node does not show the bigger picture.

Maybe a bit different approach would be better and more efficient. Some time ago I was thinking about more generalized mapping function directly from XML (something like `get_*_mapping` functions in tiff._api) just for direct copying from XML to final metadata structure. That is what I had done in bruker _api which is lots and lost of XML with loads of junk....
Oh wait, now what I touched that subject It reminded me that I have my own implementation of dictionarization function, which accepts etree XML node and returns a nice pythonic (dict, lists) tree.
get look at that: https://github.com/hyperspy/rosettasciio/blob/9bd25632c68dce920137a083227d6e8026d05610/rsciio/bruker/_api.py#L477-L501

Try your case on that an let me to know how it goes.

sem-geologist:
forgot it also needs this to work:
```python
from ast import literal_eval
```
and the function `interpret` above the `dictionarize`

nem1234:
Thank you. 
I'll try it tomorrow.
Now I'm tired.....

sem-geologist:
anyway this brings me to point to realize that there are few XML attempts and different parsing implementations currently in this library as XML is used by many vendors. @ericpre , maybe I should try unify this. mine implementation is tailored to deal with XML's made by bruker, but there are also some good universal things, like workarounds with msxml noncompliance (Some bold dude from MS once said that XML is not for interoperability, but only for local storage, thus it does not need to conform to standards... and that insane mistake was kept for decades).

nem1234:
For JEOL SightX Tiff file, the dictionarize function seems to work well with a few modifications.
It seems better than convert_xml_to_dict in tools.py.
I believe that this code is good enough for major cases.

-------------
def interpret(t):
    return(t)
--------
........
     if children:
 > #         dd = defaultdict(list)
 >         dd = {}
         for dc in map(dictionarize, children):
             for k, v in dc.items():
>                 if not (k in dd):
>                    dd[k] = []
                 dd[k].append(v)
------

In case of `<A>x<B>b1</B><B>b2</B>y<C>c<D>d</D></C>z</A>`
result is 
{'A': {'B': 'b', 'C': {'D': ['d1', 'd2', {'E': 'e'}]}}}
texts of "x", "y", "z" are ignored.
This behavior is bad for general XML decoder, but I think this kind of XML is very rare, and in such case, usually it is meaningless to convert to dictionary.

sem-geologist:
> '''def interpret(t):
> return(t)
> '''

such function wont cast values to the right type.

as for element containing both the other element and text, this SO post clarifies that quite simple:
https://stackoverflow.com/questions/66148477/can-an-xml-element-have-both-text-and-element-content

I want to quote the part of last answer made by Michael Kay:

> But people can and do use XML mixed content in other ways. Mixed content is of course the biggest difference between XML and JSON, and reflects the fact that the M in XML means "markup": it was devised as a system for labelling parts of a textual document

So my dictionarization function is for **well-formated** XML where XML which is addressed by parser is used by Bruker for metadata and data encapsulation. Looks that Bruker and Jeol uses XML differently. Bruker software stack is designed with server-communication-client paradigm, where client can be on remote machine and connect and control the detectors on server remotely through TCP/IP. Because of that XML are very strictly formatted to adhere to requirements for direct sending and receiving through TCP/IP frames. If there is any possibility for NULL bytes - it is coded with Base64. The Tiff of Jeol SightX and it s XML would brake such frame if it would be sent directly over TCP/IP without any encapsulation, and it looks for me that XML is used partly not as data structure and sanitization for network transmission, but  document form intermixed with data - in one word - A Mess (This is still not so bad, just wait for someone with newest generation Jeol SEM EDS software asking for some support - that is some breakthrough in making a mess). However if it is a document - it is **valid** XML, and rather my function (if I move it to `utils.tools`) should take this into consideration.

Another gotchas is (if You looked to how bruker parser uses it) it is not used to dictionarize whole XML, but only relevant portions (elements) skipping all the scruff behind.

sem-geologist:
I started to play myself with this, and Your xml example actually is already butchered with ElementTree.fromstring. Element tree looses `z` text. My function gets the x (which element tree catches) and places it under `#text` key under "A" values (as dictionary). So it really depends from context, maybe some parts of Jeol XML is not XML at all but rather HTML, or document like XML, it then needs a bit different parsing approach.

nem1234:
> such function wont cast values to the right type.

I know. This interpret function is just for testing. 
Basically (XML) schema is needed to decode such parameters correctly, but the schema is not officially distributed within the URI in namespace attribute.

And I did not find any markup text in JEOL SightX file. 
For this file, I don't care about markup texts.

I think that at the starting point we can use simple conversion function to analyze metadata inside image files. 
convert_xml_to_dict function in tools.py is not enough to dump JEOL SightX XML, but your code seems to work better for me.

sem-geologist:
hey @nem1234 ,
is there real life situation with your example:
```xml
<A>x<B>b1</B><B>b2</B>y<C>c<D>d</D></C>z</A>
```
I mean `x` and `z` text under tag `<A>`. I actually researched and found that it is possible to implement reading of all intertag text. The main questions are: 1) should we? 2) if yes, what ishould be returned: a) concatenated text with stripped white space (that would be `'xz'`) or b) the list of whitespace stripped strings - in that case `['x', 'z']`?
Answer would be easy to find from context if real file content with such structure would exist. As You gave an example with A,B,C,D,x,b1,b2,y,c,d,z thus I thought maybe it is based on some real life XML.
I still need to write test for #111, I probably could extend that later, but it would be nice to make it right from the beginning.

nem1234:
Currently I don't have real data with such XML elements.
(Potentially Oxford's *.oipx file (not supported yet) may contain mark-up texts to describe sample info and so on)

In my opinion,  1)	No, 
 Converted dict(list) should be easily accessible in the dict(list) / Box(BoxList) style.
     If it is really needed, it should be accessed via XML library.
     Main purpose of convert_xml_to_dict is to simplify complex XML data structure and its API.

ericpre:
https://github.com/hyperspy/rosettasciio/pull/111 has been merged, is it worth trying to start using it in this PR?

sem-geologist:
I want to give some short (not) explanation how to use that and logic behind, I started issue #113 , which I hope will have time after EMAS workshop to finish...  Basically the XML to dict translator can go through that short xml example and parse everything, if corresponding flags are set. Why it does not do that by default - because it is very possible it is junk and should be left in the dust bin of OEM XML :stuck_out_tongue:. It will depend very much from how OEM uses XML to save its data. There are few ways: painful and very painful.

ericpre (approved):
The new `XmlToDict` seems to work nicely to fix the issue with python-box >=7! Thanks @sem-geologist, @nem1234.

PR #86, opened 2023-02-07
Add (optional) code style check before commits (w/pre-commit + Black)

hakonanes:
### Description of the change
I've added an option for developers to run a code style check before commiting with git (a git hook), using [pre-commit](https://pre-commit.com/) and Black (see Black's [Version control integration page](https://black.readthedocs.io/en/stable/integrations/source_version_control.html)). These are added to the development dependencies.

The "Lint" GitHub check in this repo uses the (latest) stable version of Black to check changes. pre-commit requires specific versions of software used in their git hooks ([reasoning here](https://pre-commit.com/#using-the-latest-version-for-a-repository)). I set it to the latest, v23.1.0, however the "Lint" check and the pre-commit version might become out of sync at some point. When that happens, a pre-commit config update can be run with `pre-commit autoupdate` (and the updated config file must be pushed to `main`).

The use of this optional code style check should be added to the contributing guide. However, I noticed that in the relevant section, RosettaSciIO's contributing refers to HyperSpy's contributing guide. Should I add a brief explanation below this reference, or should all this be done in HyperSpy as well?

Also updated .gitignore to fix #85.

### Progress of the PR
- [x] add an changelog entry in the `upcoming_changes` folder (see [`upcoming_changes/README.rst`](https://github.com/hyperspy/rosettasciio/blob/main/upcoming_changes/README.rst)),
- [x] Check formatting changelog entry in the `readthedocs` doc build of this PR (link in github checks)
- [ ] Update contributing guide
- [ ] ready for review.

PR #111, opened 2023-04-27
move the xml_to_dict and msfiletime to datetime from bruker parser to utils.tools

sem-geologist:
### Description of the change
There are some dublication attempts spotted (i.e. XML handling) where different kind of
parsers have its own implementation.
This PR moves some of functions from bruker api to utils.functionaly_coresponding_file:
- move and expose  msfiletime_to_unix (datetime.datetime) in utils.date_time_tools
- prelude to unification of XML to dictionary translations: move and refactor customizable xml to dictionary translator from bruker api:
  - float sanitizer function
  - xml dictionarization

This is first step. Next step (other PR) will be format by format review if XML can be parsed using the same tools.

### Progress of the PR
- [x] msfiletime_to_unix
- [x] move xml helper tools to utils.tools, make it customizable for reusing and switch bruker api to use that.
- [x] functions contains rich docstrings (done), update user guide (after wider adoption of xml_to_dict translator to other formats in other PR)
- [x] add tests, (there is no needed, as it still do same job)
- [x] ready for review.

### DIFF
--- a/rsciio/bruker/_api.py
+++ b/rsciio/bruker/_api.py
@@ -28,7 +28,6 @@
 # ----------------------
 from os.path import splitext, basename
 from math import ceil
-import re
 import logging
 from zlib import decompress as unzip_block
 from struct import unpack as strct_unp
@@ -36,10 +35,10 @@ from datetime import datetime
 from ast import literal_eval
 import codecs
 import xml.etree.ElementTree as ET
-from collections import defaultdict
 import io
 
 from rsciio.utils.date_time_tools import msfiletime_to_unix
+from rsciio.utils.tools import sanitize_msxml_float, XmlToDict
 
 import dask.delayed as dd
 import dask.array as da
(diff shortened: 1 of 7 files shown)


sem-geologist:
Ok, this is kind of ready to review, few minor issues which I had spotted:
There is no test file for utility.functions, As I moved some function there I had to leave testing of that universalized function inside `test_bruker.py`.
Do I need to add any changelog? this does not change anything for users. This makes tools available for developers.

ericpre:
Would it make sense to write some test specific to the cases/issues it need support for? This would make future support and development much easier.

Lint is failing. For what it is worth, [pre-commit](https://pre-commit.com) configuration have been added recently in https://github.com/hyperspy/rosettasciio/pull/86.

I would add a changelog, because the changelog is also useful for developers, for example, to easily find when things have been done, the context, reasoning, etc.

sem-geologist:
@ericpre, I find current lint rules to be stupid and offensive. It asks to change better formatted (more readable) code into worse formatted (less readable)! I still use older hardware for most of my work (it is not even 2k, just old 1k laptop widescreen of good old thinkpad T420), lint complains about line formatted to fit to 79 charter and want to collapse it into single line much over 80 charters. That makes it pain to have two split files opened side-by-side. One is to allow writing longer lines (up to 120) which should be allowed exception not the rule, but forcing collapsing divided lines into single cluster***** line - that is offensive and makes it less readable - absolutely counter-productive. Who let in the Java devs setting this python lint rules?

ericpre:
Regarding the usage of `black`, I was also a bit skeptical initial for the reasons you mentioned but after using it for some time, I find it good. A lot of these things are matter of preference but it does a good job at keeping the code consistent.

What do you think about writing test specific to the cases/issues it need support for?

sem-geologist:
I think discussion section will be right place to discuss further about `black`.

I will add small made up XML and tests. I think it would be good idea to create `test_utils.py` and put it there as also move some generic tests from `test_bruker.py` there.

ericpre:
Commit [44df588](https://github.com/hyperspy/rosettasciio/pull/111/commits/44df5889ccce7ee926be88993ef576aaf8cdd1fd) can be removed now that ipython 8.13.0 have been yanked and 8.13.1 released.

sem-geologist:
thanks @ericpre , and how to do that easily?

sem-geologist:
All tests are passing, I think this is ready for review

ericpre (approved):

PR #118, opened 2023-05-16
Add support for Jeol epma data format (Issue #115)

nem1234:
### Draft
This is not suitable for merging into main stream

more sample data is needed

### Description of the change
- Minimum support for JEOL EPMA map
- Not enough metadata, filesize info
- Image size is fixed. (256x256)

### Progress of the PR
- [x] add tests,
- [ ] ready for review.

```python
from rsciio.jeol import file_reader
file_reader("your_msa_file.map")
# Your new feature...
```

This is just a minimum EPMA map support for discussion.
I don't know if new signal_type should be defined or not.

### DIFF
--- a/rsciio/jeol/_api.py
+++ b/rsciio/jeol/_api.py
@@ -153,9 +153,23 @@ def _read_asw(filename, **kwargs):
 file_reader.__doc__ %= (FILENAME_DOC, LAZY_DOC, RETURNS_DOC)
 
 def _read_epma_img(filename, **kwargs):
+    """
+    Parameters
+    ----------
+    filename : str
+        img file name
+    kwargs :
+        not used
+
+    Returns
+    -------
+    image_list : list of images(dict)
+        (Always len(image_list) == 1)
+        None if image is not an EPMA map.
+    """
     with open(filename, "br") as fd:
         file_magic = np.fromfile(fd, "<I", 1)[0]
-        if file_magic != 1:
+        if file_magic != 1:   #  Is this correct?
             return None
         fd.seek(32)
         _w = 256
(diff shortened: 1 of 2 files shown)


nem1234:
Thank you.
I recognized that this imprimentation is very poor, and does not cover whole JEOL EPMA files.
The file format seems different to current JEOL's TEM EDS format,
This file seems a "mapping", not a fullly colleced raw data set.

In my recognization, PR should be small, step-by-step commit to main tree.
In this stage, I only implement on Issue #115.
If I should not implement such limited purpose code, I'll retract this PR.

CSSFrancis:
@nem1234 you can always open a PR and set the PR as a draft! That helps to signal that you are still working on some things or that you might need help.

Getting feedback is a part of the process and it is nice to know what people are doing so things aren't duplicated! My general rule of thumb is try and get something working to the point where I can at least provide an example of an implementation and then submit a PR. 

If you want you can reopen this PR and mark it as a draft then maybe we can think about generalization of the file format if we can find more methods.

@sem-geologist I'd say that more of these processed datasets do have a place in rosettasciio.  If the idea is to reduce the dependency on 3rd party applications I think that's pretty well in line but we could open an issue or discussion.

sem-geologist:
wait, is it not possible to get some more data from the author of this file. I think this could be good to have. As I said, during EMAS workshop I had seen really relevant presentations utilizing HyperSpy behind. Multi-WDS instruments (basically one of optional characteristics of EPMA instrument) can produce more concentrated view on X-ray generated in the sample. As I said I see potential of all these advanced methods built inside hyperspy, where instead of EDS spectral image, stacked raw or net count WDS cube could be used. Blind source separation, PCA and other methods should work same and maybe even better than on EDS, as this would be much less noisy measurements and much more sensitive to low concentrations. While HyperSpy lacks on X-ray quantification, I got convinced that not always there is a need or even is possible to use standard based quantification. So please don't close, this would motivate me to bring in Cameca formats here.

PR #121, opened 2023-05-19
expose some functions under bruker.api

sem-geologist:
### Description of the change
This PR aims to modernise and extend the `bruker._api.py`:
- expose middle layer functions and classes under `bruker.api`
- review, update and extend the docstrings, in particular of these functions and class'es which or which child are
  going to be exposed through `.api`
- review the code for possible cleanup, scope-separation and streamlining for public exposition in `.api`

### Progress of the PR
- [x] expose `SfsReader`
- [ ] expose `xml_to_spectrum`
- [ ] expose `xml_to_image` 
- [ ] add tests,
- [ ] ready for review.

???
are `xml_to_spectrum` and `xml_to_image` good naming for these functions. functions require particular etree nodes - maybe `et_node_to_spectrum` and `et_node_to_image` would be better naming?

```python
from sciio.bruker import api as b_api

b_api.SFSReader('somefile.pan')  # *.pan are particle analysis files using same container as bcf 
```

### DIFF
--- a/rsciio/bruker/_api.py
+++ b/rsciio/bruker/_api.py
@@ -47,7 +47,7 @@ import numpy as np
 from rsciio.docstrings import FILENAME_DOC, LAZY_DOC, RETURNS_DOC
 
 __all__ = [
-    "SfsReader",
+    "SFSReader",
 ]
 
 _logger = logging.getLogger(__name__)
@@ -277,25 +277,21 @@ class SFSTreeItem(object):
         return data
 
 
-class SfsReader(object):
-    """Class to read sfs file. SFS is known to be used by bruker for
-    saving particle analysis (*.pan) frames, hyperspectral maps and
+class SFSReader(object):
+    """Class to read sfs file. SFS is known to be used by Bruker for
+    saving particle analysis (*.pan) frames, hyper-spectral maps and
     EBSD maps and metadata (*.bcf).
     SFS stands for AidAim software's(tm) [s]ingle [f]ile [s]ystem.
     The class provides basic reading capabilities of such container.
     It is capable to read compressed data in zlib, but
-    SFS can contain other compression which is not implemented here.
-    It is also not able to read encrypted sfs containers.
+    SFS can use other compression than zlib, which is not implemented
+    here. This implementation also does not cover SFS internal encryption.
 
-    This class can be used stand alone or inherited in construction of
+    This class can be used stand-alone or inherited in construction of
     file readers for files using this format.
 
     The format can be recognised independently from file extension
     by file "magic" in the first 8 bytes: AAMVHFSS
-
-    Attributes
-    ----------
-    filename
 
     Attributes:
     ----------
@@ -313,7 +309,7 @@ class SfsReader(object):
     n_tree_items - number of items in tree descriptor (including files and
       directories)
     sfs_n_of_chunks - total number of blocks in the SFS file (serves
-      no purpoise in reading, probably used for appending data)
+      no purpose in reading, probably used for appending data)
 
     Methods:
     ----------
@@ -322,6 +318,8 @@ class SfsReader(object):
     """
 
     def __init__(self, filename):
+        """SFSReader requires valid relative or absolute path-like object.
+        which is used in python built-in open() function"""
         self.filename = filename
         self.compression = None
         # read the file header
@@ -407,11 +405,12 @@ class SfsReader(object):
         if aacs == 0x53434141:  # AACS as string
             item.uncompressed_block_size = uc_size
             item.n_compressed_blocks = n_of_blocks
-        else:
+        else:  # pragma: no cover
             raise ValueError(
                 """The file is marked to be compressed,
 but compression signature is missing in the header. Aborting...."""
             )
+            # Such kind of bug failed to be replicated more than once
 
     @staticmethod
     def _flat_items_to_dict(paths, temp_item_list):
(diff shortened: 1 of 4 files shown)


rsciio/bruker/_api.py
```diff
@@ -289,25 +277,53 @@ def get_as_BytesIO_string(self):
...
     This class can be used stand alone or inherited in construction of
-    file readers using sfs technolgy.
+    file readers for files using this format.
+
+    The format can be recognised independently from file extension
+    by file "magic" in the first 8 bytes: AAMVHFSS
 
     Attributes
     ----------
     filename
```
jlaehne, line 298: duplicated below

rsciio/bruker/_api.py
```diff
@@ -289,25 +277,53 @@ def get_as_BytesIO_string(self):
...
+    sfs_version - SFS version (for debug)
+    chunksize - block size including block header and data (often a "round"
+      number in hexidecimali, but in EBSD cases could be "optimized")
+    usable_chunk - usable size within single block after subtraction size
+      of block header (-32 bytes)
+    tree_address - block index where file tree descriptor starts
+    n_tree_items - number of items in tree descriptor (including files and
+      directories)
+    sfs_n_of_chunks - total number of blocks in the SFS file (serves
+      no purpoise in reading, probably used for appending data)
```
jlaehne, line 316:
```suggestion
      no purpose in reading, probably used for appending data)
```

sem-geologist:
@ericpre , @jlaehne I have some question. As I am doing pretty extensive review of bruker _api.py, so that it could be more streamlined, I came again over idea how to handle images. Bruker xml image node can have plains, it is basically video channels, which registers signal at same time when beam is rastered over sample ROI. So basically most of metadata is the same except description string of the channel. Hitherto in this case all metadata was being duplicated, and every plain was returned as separate independent set with its data and metadata. But... is it the right and most efficient way? would it not be better way to stack those plains, alongside axis called as "channels"? can axes have labeled scale (that is instead of numbers it would be channel description)? Does this makes any sense at all? I mean for me it does a lot, as those plains are different representations of exactly same material interaction or result of  beam: i.e. say as example 4 channels: BF, DF, DF4, HAADF. They all represent different interacted electrons with matter.
Even if taking 2 channel (most common setup with Bruker) of BSE and SE images, those represents different energy electrons, and in some sense combining it forms kind of spectral image.

such change probably  is going break some established workflows. But I think it would be very right thing to do.

jlaehne:
We had a similar topic brought up for LumiSpy (CL spectral image + corresponding SE image) by @jordiferrero, which was never implemented so far: https://github.com/LumiSpy/lumispy/issues/73
Indeed, it would be nice to have an upstream solution for that in HyperSpy!

Concerning the 'labeled' axes for stacked data, @CSSFrancis was working on a PR.

CSSFrancis:
So adding labeled axes wouldn't be terribly hard in hyperspy. I have done most of the work in https://github.com/hyperspy/hyperspy/pull/3031 Right now there is a little bit of hesitation with this change but I think that there isn't really anything stopping this from happening.   This might be something worth voting on or having a larger discussion about how these signals are handled. 

Handling multiple signals is a little bit tricky and becomes more tricky if they have different numbers of pixels, but even that can be handled by hyperspy.

There are a couple of things:

1. If all of the signals have the same dimensionality and same size --> pass as labeled axes and create a stacked array
2.  If all of the signals have the same dimensionality and different --> pass as labeled axes and create a stacked ragged array
3. If one of the signals has a higher dimensionality (i.e. HAADF and 4-D STEM) pass one as navigator and one as array

For 3 I also think that in the case where you have multiple different signals you should be able to pass all of them.  In that case you could have multiple different navigators.  For example with a 4-D STEM you could have brightfield and darkfield images.

sem-geologist:
thanks @CSSFrancis , Your 3 point division makes it pretty clear. So "planes" used in bruker XML image actually not only fulfills point 1. of same dimentiality and resolution/size, but are even more closely related, as it have exactly same column conditions, and is pixel-to-pixel generated with exactly same beam simultaneously - thus even more - it should be stacked. Hopefully Your PR https://github.com/hyperspy/hyperspy/pull/3031 will get accepted. I probably will stack images with unitless axes as for now.

Hyperspy is not even my main/only aim with this streamlining attempt. I want this to stay same useful for Hyperspy 2.0 (actually be more useful for extensions, with EBSD, XRF... in mind), but also to pave easier road for my own software (HussariX).

CSSFrancis:
> Hyperspy is not even my main/only aim with this streamlining attempt. I want this to stay same useful for Hyperspy 2.0 (actually be more useful for extensions, with EBSD, XRF... in mind, which I plan to address in some following PR's), but also to pave easier road for my own software (HussariX).

In these types of situations "what is good for the goose is good for the gander" or both software packages will probably benefit from a consistent approach to these low level problems :)  It also helps with interoperability which can only be a good thing. 

To be honest the case #2 above is quite difficult from the perspective of hyperspy.  There is nothing that says hyperspy cannot have a signal of signals similar to how `numpy` can have a ragged array of arrays. I considered something like that [here](https://github.com/hyperspy/hyperspy/issues/3078) but I am not sure if there is a consistent way to approach that.  For now there hasn't really been a real need for it so I haven't really paid it much attention.

ericpre:
Just for me to make sure that I don't misunderstand, @sem-geologist what you are asking is to have a signal, which contains BSE and SE images stacked together and another one which contains the EBSD or EDS data?
Could it be easily stack after loading the data by the library itself, at a stage, where it is known how the data needs to be structure/handle. I am not sure that this is the responsibility of RosettaSciIO to define this?

Currently, in hyperspy, these are considered as treated as different dataset and they expected to be handle at the workflow level (script, notebook, library etc.). Maybe what RosettaSciIO could do is to assign some dataset to a specify category: "main", "auxiliary" (data acquired simultaneously but not "main", or already processed data, i.e. map from a spectrum image), "survey", etc.

PR #123, opened 2023-05-24
Remove packaging of test data and download them on demand

ericpre:
I had a go at removing the packaging of the test file and downloading on demand when running the test suite.

Summary of the approach:
- test data live in this repository, just a different folder
- adding test data during development is similar as before
  - add the file to the git repository
  - update `registry.txt` by calling the `rsciio.tests.registry_utils.update_registry` function
- the test data are not packaged
- the test data are downloaded automatically if there are not locally available when running the test suite. It slow down running because:
  - it takes about 3 min to download everything because there are many small files and the overhead is significant.
  - running several worker is possible, which I suspect is a bug `pytest-xdist` or `pytest`

Example of the download all test data before running the test suite: https://github.com/hyperspy/rosettasciio/actions/runs/5071177296/jobs/9107256002?pr=123

### Drawback
- Running the test suite will fail, it the user doesn't write access to the test folder.
- There is an issue with generating hash on windows of text, xml file (most likely related to line ending) and it doesn't seem that we can solve it. However, we would workaround with a custom pre-commit hook that we could run from the pull request and that will be skipped on windows.

### Progress of the PR
- [x] Move test data a the `rsciio/tests/data` folder,
- [ ] update contributing `CONTRIBUTING.rst` once we are happy with the approach
- [ ] Check setting base url plays well with our setup in rosettasciio to pick tagged or development branch
- [x] add tests,
- [x] ready for review.

### After the PR is merged
- [ ] Update base url from my fork to the hyperspy organisation.
- [ ] Add custom pre-commit hook to update the registry so that the registry can be updated by running pre-commit from the PR (add a comment)
- [ ] follow up with the zip file


rsciio/tests/test_bruker.py
```diff
@@ -21,35 +21,37 @@
 np_file = ["30x30_16bit.npy", "30x30_16bit_ds.npy"]
 spx_files = ["extracted_from_bcf.spx", "bruker_nano.spx"]
 
-my_path = os.path.dirname(__file__)
+TEST_DATA_DIR = Path(__file__).parent / "data" / "bruker"
+FILE1 = TEST_DATA_DIR / "test1.blo"
+FILE2 = TEST_DATA_DIR / "test2.blo"
```
jlaehne, line 26:
```suggestion

```

pasted from blockfile

rsciio/tests/test_emd.py
```diff
@@ -21,13 +21,15 @@
 # National Lab (see https://emdatasets.com/ for more information).
 # NOT to be confused with the FEI EMD format which was developed later.
 
+import os
```
jlaehne, line 24:
```suggestion
```

as far as i can see, `os` is not used any more

rsciio/tests/test_tiff.py
```diff
@@ -37,23 +37,21 @@
...
 
-MY_PATH = os.path.dirname(__file__)
-MY_PATH2 = os.path.join(MY_PATH, "tiff_files")
+TEST_DATA_PATH = Path(__file__).parent / "data" / "tiff"
+TEST_NPZ_DATA_PATH = Path(__file__).parent / "data" / "npz"
 TMP_DIR = tempfile.TemporaryDirectory()
 
 
 def teardown_module():
     TMP_DIR.cleanup()
```
jlaehne, line 46: `TMP_DIR` still needed as `tmp_path` is now used?

rsciio/tests/test_edax.py
```diff
@@ -29,9 +29,8 @@
 hs = pytest.importorskip("hyperspy.api", reason="hyperspy not installed")
 requests = pytest.importorskip("requests", reason="requests not installed")
 
-
-MY_PATH = os.path.dirname(__file__)
-ZIPF = os.path.join(MY_PATH, "edax_files.zip")
+TEST_DATA_PATH = Path(__file__).parent / "data" / "edax"
+ZIPF = TEST_DATA_PATH / "edax_files.zip"
 TMP_DIR = tempfile.TemporaryDirectory()
```
jlaehne, line 34: use `tmp_path` here as for other plugins?

CSSFrancis:
@ericpre Very cool!  I'll probably try to check this out tomorrow to see if I can get it to work.  The `rsciio.tests.registry_utils.update_registry` function seems reasonable but I wonder if people will get stuck on that.  Is it worth it to see if we can use pre-commit to get that to check for new files and then automatically edit the registry.txt file?

rsciio/tests/test_edax.py
sem-geologist: maybe it is good chance to get rid of zip and other packed data?
Why:
1) bandwitch saving is redundant, as git compresses data behind anyway.
2) size is moot point, if this is going to be downloaded on demand
3) updating with new files, requires huge binary differences, getting rid of zip/gz collections, only new file will be added as difference to the git history.
4) finally, these tmp_path would be not needed (albeit if tests would be rewritten to use rsciio to read the data explicitly, and retrieved list/dictionary fed to hyperspy to generate signal for testing, for some formats loading could be straight from zip file, without any tmp_folders)

ericpre:
> @ericpre Very cool! I'll probably try to check this out tomorrow to see if I can get it to work. The `rsciio.tests.registry_utils.update_registry` function seems reasonable but I wonder if people will get stuck on that. Is it worth it to see if we can use pre-commit to get that to check for new files and then automatically edit the registry.txt file?

Yes, this is a good idea, even if not everyone uses pre-commit, it will be useful to request pre-commit to update it from a comment in a PR.
In any case, if the registry is out of date, the `Build / Test Packaging` will fail, so it will not fall through the cracks!

rsciio/tests/test_edax.py
ericpre: Yes, we should try to remove of the zip files.

Just to avoid misunderstand, the expected usage is to download the files on demand when running from a non-development installation. For developer, there will be no change in workflow (well, you can remove the data folder and it will be downloaded again).
The main aim is to be able to run the test suite from the conda-forge feedstock or the hyperspy-bundle.

> 4. finally, these tmp_path would be not needed (albeit if tests would be rewritten to use rsciio to read the data explicitly, and retrieved list/dictionary fed to hyperspy to generate signal for testing, for some formats loading could be straight from zip file, without any tmp_folders)

I am not sure about that. What would be the benefit? For development purpose I usually prefer to open the file directly, sometimes using different approach, etc.

sem-geologist: if zip files are get rid (replaced with unpacked content), then no tmp_path is needed as all files can be just opened directly (using Hyperspy `load`). That is prefered way.

Sorry for convoluted message. What I had meant some of readers (undocumented behavior) are able to open not only physical files, but can accept the opened file or other `BytesIO` like opened object, i.e. file from within zip archive - no tmp folder then is needed, as byte-stream is read and parsed directly seamlessly from zip. I know that that works with tiffs. Ofc for test files that feature rather not good due to 1,2,3 reasons given before. However in the future we should consider documenting this behavior as gz, zip are one of most common available well documented, tested, widely supported containers. I.e. it is nice to be able to keep whole SEM session for one sample within single file.

CSSFrancis:
> Yes, this is a good idea, even if not everyone uses pre-commit, it will be useful to request pre-commit to update it from a comment in a PR. In any case, if the registry is out of date, the `Build / Test Packaging` will fail, so it will not fall through the cracks!
> 
> Can we leave this for another PR? :)

Sounds good to me.  Otherwise I think it looks good.

rsciio/tests/test_edax.py
ericpre: Ignoring the zip file business, for the scenario where `tmp_path` are used for testing saving file, we should keep this pattern, because we don't need to worry about removing them, user having write access to the folder, etc. What do you think?

This make me think that to run the test suite, the user will need write access to the tests folder, which is not guaranty in multi user installation. I have updated the summary of this PR.

rsciio/tests/test_emd.py
ericpre: It is still line 124.

rsciio/tests/test_tiff.py
ericpre: Yes, because with unzipping file it is not convenient to use the pytest fixture

rsciio/tests/test_edax.py
ericpre: > use `tmp_path` here as for other plugins?

Same as above.

sem-geologist: of course for saving tests the situation is different, and rather it can't be evaded anyhow.

rsciio/tests/test_emd.py
jlaehne: Couldn't the same as in line 185 be used in line 124:
`hdf5_dataset_path = tmp_path / "test_dataset.emd"`

jlaehne:
TODO before merging: Edit `CONTRIBUTING.rst` to update the path where test files should be placed (and if needed how to update the registry). The doc currently contains the comment *[change for pooch]* at the appropriate point.

rsciio/tests/test_emd.py
ericpre: Using that pattern doesn't work with the `setup_method` and the temporary director is required for the whole lifetime of  test class. They may be other way, but I didn't spend much effort on this, as this is secondary in this PR.

jlaehne: OK, just thought we could get things working more similar with little effort

CONTRIBUTING.rst
```diff
@@ -39,6 +39,25 @@ source code or by using `pre-commit <https://pre-commit.com>`_ to format code au
 Alternatively, the comment ``pre-commit.ci autofix`` can be added to a PR to fix the formatting
 using `pre-commit.ci <https://pre-commit.ci>`_.
 
+Adding and Updating Test Data
+-----------------------------
+The test data are located in the corresponding subfolder of the ``rsciio/tests/data`` folder.
+To adding or update test data:
```
jlaehne, line 45:
```suggestion
To add or update test data:
```

CONTRIBUTING.rst
```diff
@@ -39,6 +39,25 @@ source code or by using `pre-commit <https://pre-commit.com>`_ to format code au
...
+
+#. use git as usual to add file to the repository.
```
jlaehne, line 47:
```suggestion
#. use git as usual to add files to the repository.
```

CONTRIBUTING.rst
```diff
@@ -39,6 +39,25 @@ source code or by using `pre-commit <https://pre-commit.com>`_ to format code au
...
+#. Update ``rsciio.tests.registry.txt``.  The test data are not packaged in RosettaSciIO to
+   keep the packages as small as possible in size. However, to be able to run the test suite
+   of RosettaSciIO after installation or when packaging on conda-forge, pooch is used to
+   download the data when necessary. It means that when adding and updating test file, it
```
jlaehne, line 51:
```suggestion
   download the data when necessary. It means that when adding and updating test files, it
```

CONTRIBUTING.rst
```diff
@@ -39,6 +39,25 @@ source code or by using `pre-commit <https://pre-commit.com>`_ to format code au
...
+   is necessary to update the registry ``rsciio.tests.registry.txt`` which can be done by
```
jlaehne, line 52:
```suggestion
   is necessary to update the registry ``rsciio.tests.registry.txt``, which can be done by
```

CONTRIBUTING.rst
```diff
@@ -98,7 +117,9 @@ directory should contain the following files:
 **Tests** covering the functionality of the plugin should be added to the
 ``tests`` directory with the naming ``test_spamandeggs.py`` corresponsing to
 the plugin residing in the directory ``spamandeggs``. Data files for the tests
-should be placed in a corresponding subdirectory [change for pooch].
+should be placed in a corresponding subdirectory - see
```
jlaehne, line 120:
```suggestion
should be placed in a corresponding subdirectory - see the
```

CONTRIBUTING.rst
```diff
@@ -98,7 +117,9 @@ directory should contain the following files:
...
+[Adding and Updating Test Data](#adding-and-updating-test-data) section for more
```
jlaehne, line 121: Link not working in the doc-build of this PR: https://rosettasciio--123.org.readthedocs.build/en/123/contributing.html

ericpre: Markdown syntax in a `rst` file! 🤦‍♂️

rsciio/conftest.py
```diff
@@ -28,3 +31,15 @@
...
+
+
+def pytest_configure(config):
+    # Run in pytest_configure hook to avoid capturing stdout by pytest and
+    # inform user that the test data are being downloaded
+
+    # Workaround to avoid running it for each worker
+    worker_id = os.environ.get("PYTEST_XDIST_WORKER")
+    if worker_id is None:
+        print("Checking if test data need downloading...")
```
jlaehne, line 43:
```suggestion
        print("Checking if test data needs to be downloaded ...")
```

rsciio/tests/registry_utils.py
```diff
@@ -0,0 +1,148 @@
...
+    with open(output, "w") as outfile:
+        for fname, fhash in zip(files, hashes):
+            # Only use Unix separators for the registry so that we don't go
+            # insane dealing with file paths.
+            outfile.write("'{}' {}\n".format(fname.replace("\\", "/"), fhash))
+
+
+def download_all(pooch_object=None, ignore_hash=None, progressbar=True):
+    """
+    Download all test data if there are not already locally available in
```
jlaehne, line 107:
```suggestion
    Download all test data if they are not already locally available in the
```

jlaehne (approved):

ericpre:
Everyone is happy with it, merging to continue on the other PRs.

PR #124, opened 2023-05-25
Add support for pixelwise EDS live time / real time into digital micrograph file reader

nem1234:
# Draft
Not ready to merge

# Description of the change
- Add suport for pixel-wise live time / real time for EDS image
- Should these maps be converted to images or not?

## ToDo:
- sum / average of pixel-wise live-time and real-time.
-  What is EDS.Acquisition.Live-time multiplier ?

### Progress of the PR
- [ ] add tests,
- [ ] ready for review.

```python
import hyperspy.api as hs
s = hs.load("eds-si.dm4")
print(s.metadata.EDS.Real_time_map)
print(s.metadata.EDS.Live_time_map)
print(s.metadata.EDS.Count_rate_map)
```

### DIFF
--- a/rsciio/digitalmicrograph/_api.py
+++ b/rsciio/digitalmicrograph/_api.py
@@ -1064,6 +1064,18 @@ class ImageObject(object):
                         "Acquisition_instrument.TEM.Detector.EDS.solid_angle",
                         None,
                     ),
+                    "{}.EDS.Images.Count rate".format(tags_path): (
+                        "EDS.Count_rate_map",
+                        lambda x: np.array(x).reshape(self.shape[1:]),
+                    ),
+                    "{}.EDS.Images.Live time".format(tags_path): (
+                        "EDS.Live_time_map",
+                        lambda x: np.array(x).reshape(self.shape[1:])
+                    ),
+                    "{}.EDS.Images.Real time".format(tags_path): (
+                        "EDS.Real_time_map",
+                        lambda x: np.array(x).reshape(self.shape[1:])
+                    ),
                     "{}.EDS.Live_time".format(tags_path): (
                         "Acquisition_instrument.TEM.Detector.EDS.live_time",
                         None,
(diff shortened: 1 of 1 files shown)


sem-geologist:
I am really interested into outcomes of this PR. During EMAS workshop I had an opportunity to talk with few different engineers from different vendors of EDS. Different vendors have different approaches to the same problem. In case of Bruker, live time is encoded inside EDS spectra (known as 0eV peak). Just as a funny thing, from one of other vendor representative when discussing how dead times are handled in other vendor EDS, 0eV peak was commented by one as the most brilliant way, and other as most ******-way (Both engineers was not from Bruker, just to clarify).

So for some time I was wondering if live_time should be returned as static array, or as a function, or property (which would trigger catched getter... @nem1234  what you would suggest? Also how to deal with fact that Hyperspy expects a single value instead of array as live time?

nem1234:
Umm..... 

We want to compare the quantity of elements between EDS and EELS.
In such case, pixel time is very important and want to correct the signal intensity using pixel live time.
But, most users just want a mapping of element, and they don't care pixel-wise live-time.

And in our case, (may be depend on the combination of TEM and EDS), overall live-time is not recorded into digital micrograph tags.
To calculate overall live/dead time, we need pixel-wise live time table.

In my opinion,  
   "static array, or as a function, or property"  these are good enough, but should be accessible via common method not depend on equipments / io plugins.



My current problems....
1. real_time is not same as pixel time setting. Pixel time in metadata should be overwritten with real/live time?
2. Only one live time map is recorded when multi cycle scanning is selected.
3. Overall live/dead-time is not recorded.  They need to be calculated from pixel-wise data.
4. real/live_time is fluctuated in 20-30% of setting value when software sync is selected.

Hyperspy related problems:
1. In hyperspy/io.py, post_process functions run before mapping functions.
    Calculation of overall live-time can not be placed into post_process.  
2.    Mapping and post_process functions should be placed in rosettasciio, 
   they should not depends on hyperspy.

PR #132, opened 2023-06-03
Tidy up namespace

ericpre:
To avoid having the module (`docstrings`, `exceptions`, `conftest`) when using autocompletion, etc.

### Progress of the PR
- [x] Make `docstrings` module private,
- [x] Move `exceptions` to `utils` folder,
- [x] Move `conftest.py` to `tests` folder,
- [n/a] add tests,
- [x] ready for review.


sem-geologist:
just why? Is this java project? While I understand that docstrings could be "private", I don't grasp any benefits in hiding exceptions. So if Someone will want to catch the rsciio specific expection he will need to import private model? This is so wrong on many levels...

upcoming_changes/132.maintenance.rst
```diff
@@ -0,0 +1 @@
+Tidy up ``rsccio`` namespace: privitise ``docstings``, ``exceptions`` and move ``conftest.py`` to tests folder.
```
jlaehne, line 1:
```suggestion
Tidy up ``rsciio`` namespace: privatise ``docstrings``, ``exceptions`` and move ``conftest.py`` to tests folder
```

ericpre:
> just why? Is this java project? While I understand that docstrings could be "private", I don't grasp any benefits in hiding exceptions. So if Someone will want to catch the rsciio-specific exception he will need to import private module? This is so wrong on many levels...

Alright @sem-geologist, agreed, it makes sense to keep `exceptions` public, it is now moved to the utils folder.

At some point, in another PR, we should review the content of the utils folder and define what makes to keep as public and document accordingly explicitly in the API reference. At the moment, there is nothing about it in the API reference.

upcoming_changes/132.maintenance.rst
```diff
@@ -0,0 +1 @@
+Tidy up ``rsciio`` namespace: privatise ``docstings``, move ``conftest.py`` and ``exceptions`` to tests and utils folder, respectively
```
jlaehne, line 1:
```suggestion
Tidy up ``rsciio`` namespace: privatise common ``docstrings``, move ``conftest.py`` and ``exceptions`` to tests and utils folder, respectively
```

jlaehne: still stings and not strings :-)

jlaehne (approved):

jlaehne:
@sem-geologist if you agree with the updated version of this PR, I think we could merge it

sem-geologist:
@ericpre , yes its ok

ericpre:
Thanks @sem-geologist, @jlaehne!

PR #162, opened 2023-09-25
Load `.mrc` using Distributed and with Direct Electron Metadata

CSSFrancis:
### Description of the change
This adds support for loading .mrc files using the `dask-distributed` backend and memory mapping.  This is helpful for  loading large files using multiple threads split across many machines.

The methods in the `_distributed` file should be fairly extendable and can be applied to other binary files which use the `np.memmap` function.

Changes:
----------
- Added `memmap_distributed` function 
- Add load metadata.txt file from DirectElectron
- Added `distributed` and `chunks` kwargs to the .mrc file loader

### Progress of the PR
   - [x] Add `memmap_distributed` function 
   - [x] Add load metadata.txt file from DirectElectron
- [ ] add tests
   - [x] Test loading using distributed option
   - [ ] Test using metadata.txt file (Need to make small test files with different sizes)
- [ ] ready for review.

```python
from rsciio.mrc import file_reader
file_reader("your_mrc_file.msa", 
                    distributed=True,  # can use `dask-distributed` if wanted 
                    lazy=True, # If lazy= False will still use distributed loading but then compute
                    metadata_file = "de_metadata.txt",  # Direct Electron saved  metadata file 
                    )
```


CSSFrancis:
pre-commit.ci autofix

CSSFrancis:
@ericpre so it seems like the Packaging CI seems to fail when adding a new file.  Is that correct?  It seems like it is trying to pull the file from the registered list to package (and then subsequently prune).  Is this something worth trying to fix? Or just wait until the file is merged into main.

ericpre:
Yes, the packaging failure is normal, it will pass once the PR is merged.

rsciio/utils/distributed.py
```diff
@@ -0,0 +1,169 @@
...
+    file,
+    dtype,
+    offset=0,
+    shape=None,
+    order="C",
+    chunks="auto",
+    block_size_limit=None,
+):
+    """
+    Drop in replacement for py:func:`numpy.memmap` allowing for distributed loading of data.
```
ericpre, line 117:
```suggestion
    Drop in replacement for :class:`numpy.memmap` allowing for distributed loading of data.
```

rsciio/utils/distributed.py
```diff
@@ -0,0 +1,169 @@
...
+
+    This always loads the data using dask which can be beneficial in many cases, but
+    may not be ideal in others. The chunks and block_size_limit are for describing an ideal chunk shape and size
```
ericpre, line 120:
```suggestion
    may not be ideal in others. The ``chunks`` and ``block_size_limit`` arguments are for describing an ideal chunk shape and size
```

rsciio/utils/distributed.py
```diff
@@ -0,0 +1,169 @@
...
+    chunks : tuple, optional
+        Chunk shape. The default is "auto".
+    block_size_limit : int, optional
+        Maximum size of a block in bytes. The default is None.
+    dtype : numpy.dtype, optional
+        Data type. The default is None.
+
+    Returns
+    -------
+    dask.array.Array
```
ericpre, line 50: In `Returns`, the docstring for `chunks` is missing

rsciio/utils/distributed.py
```diff
@@ -0,0 +1,169 @@
...
+
+import numpy as np
+import dask.array as da
+
+
+def get_chunk_slice(
+    shape,
+    chunks="auto",
+    block_size_limit=None,
+    dtype=None,
```
ericpre, line 28: Would it be worth adding tests against a some possible values of `shape` and `chunks`?

rsciio/utils/distributed.py
```diff
@@ -0,0 +1,169 @@
...
+    Parameters
+    ----------
+    shape : tuple
+        Shape of the data.
+    chunks : tuple, optional
+        Chunk shape. The default is "auto".
+    block_size_limit : int, optional
+        Maximum size of a block in bytes. The default is None.
+    dtype : numpy.dtype, optional
+        Data type. The default is None.
```
ericpre, line 46: Specify that these two are used by `da.core.normalize_chunks` when `chunks="auto"`

rsciio/utils/distributed.py
```diff
@@ -0,0 +1,169 @@
...
+    )
+    for ind in np.ndindex(chunks_shape):
+        current_chunk = [chunk[i] for i, chunk in zip(ind, chunks)]
+        starts = [int(np.sum(chunk[:i])) for i, chunk in zip(ind, chunks)]
+        stops = [s + c for s, c in zip(starts, current_chunk)]
+        slices[ind] = [[start, stop] for start, stop in zip(starts, stops)]
+    return da.from_array(slices, chunks=(1,) * len(shape) + slices.shape[-2:]), chunks
+
+
+def slice_memmap(sl, file, dtypes, shape, **kwargs):
```
ericpre, line 74: Avoid using an abbreviation?
```suggestion
def slice_memmap(slices, file, dtypes, shape, **kwargs):
```

rsciio/utils/distributed.py
```diff
@@ -0,0 +1,169 @@
...
+    """
+    Slice a memmaped file using a tuple of slices.
+
+    This is useful for loading data from a memmaped file in a distributed manner. This takes
+    a slice of the dimensions of the data and returns the data from the memmaped file sliced.
+
+    Parameters
+    ----------
+    sl : array-like
```
ericpre, line 83:
```suggestion
    sl : array-like of int
```

rsciio/utils/distributed.py
```diff
@@ -0,0 +1,169 @@
...
+        An array of the slices to use. The dimensions of the array should be
+        (n,2) where n is the number of dimensions of the data. The first column
+        is the start of the slice and the second column is the stop of the slice.
+    file : str
+        Path to the file.
+    dtypes : numpy.dtype
```
ericpre, line 89: Why is `dtypes` plural?

rsciio/utils/distributed.py
```diff
@@ -0,0 +1,169 @@
...
+        Data type of the data for memmap function.
```
ericpre, line 90: This is the shape of the whole dataset? From reading the code, I would have expected that this would be the shape of the slice?

rsciio/mrc/_api.py
```diff
@@ -136,8 +141,118 @@ def get_data_type(mode):
...
+                axes_dict = {
+                    "name": axes_names[i],
+                    "size": s,
+                    "units": axes_units[i],
+                    "navigate": True,
+                    "index_in_array": ind,
+                }
+                if axes_scales[i] != -1:  # -1 means that the scale is not defined
+                    axes_dict["scale"] = axes_scales[i]
+                else:
```
ericpre, line 218: Not sure, this is worth adding a test for this?
```suggestion
                else:  # pragma: no cover
```

rsciio/utils/distributed.py
CSSFrancis: It's a result of the `map_blocks` function also having a dtype parameter which causes some problems.

rsciio/utils/distributed.py
CSSFrancis: This is the shape of the entire dataset.  You first need to create the entire memory mapped dataset and then only slice the parts that you want to build a single chunk.  

I've tried a couple of other things such as just reading the relevant lines and they are pretty much equivalent all things considered.  Plus the syntax is a little more consistent this way.

rsciio/utils/distributed.py
CSSFrancis: Thats a good idea.  I've kind of gotten into the habit of using sl or slic just because ``slice`` is protected but slices is probably the better thing to use.

rsciio/utils/distributed.py
CSSFrancis: Huh.. I didn't realize that was a class.  Numpy gets a little weird with not capitalizing their classes...

rsciio/utils/distributed.py
CSSFrancis: In most cases yes but I think that the `map_blocks` function pops off the `dtype` argument.  So if you want to specify a `dtype` you have to give it a different name otherwise dask will eat that kwarg. 

There are also cases where you might have a different dtype. For example some data is saved with an array and then a timestamp and then another array.  In these cases it might be better to have a custom dtype which accounts for these breaks.  I was going to add that to this function eventually.

CSSFrancis: In any case the user shouldn't really be using the `slice_memmap` function directly and should instead be using the `memmap_distributed` function which copies the syntax for the `memmap` function.

ericpre: Yes, you are right! 👍

upcoming_changes/162.new.rst
```diff
@@ -0,0 +1,5 @@
+Added the :func:`~rsciio.utils.distributed.memmap_distributed` function for loading a memmap file
+from multiple processes.
+
+Added the arguments ``distributed`` and ``metadata_file`` to the .mrc file reader for loading metadata
+save from DirectElectron detectors.
```
ericpre, line 5: Is it worth mentioning the speed improvement (remove reshape, transpose) from the last commit?

CSSFrancis: Done!

ericpre (approved):

PR #164, opened 2023-09-28
Update of HyperSpy Markers API changes for the `hspy`/`zspy` format

ericpre:
Follow up of https://github.com/hyperspy/hyperspy/pull/3148.

### Progress of the PR
- [x] Update to new markers API,
- [x] fix parsing tuple of string,
- [x] bump matploltib minimum requirement to support `matplotlib.collections.Collection.set_offset_transform` which was added in matplotlib 3.5 
- [x] Various fixes dealing with ragged array
  - [x] `ragged` attributes wasn't set properly in `save`/`load` cycle
  - [x] dtype of ragged array was incorrectly saved in `zspy`
  - [x] Fix issue with several ragged arrays dataset in the same group
- [x] update user guide: bump file version,
- [x] add tests,
- [x] ready for review.

### DIFF
--- a/rsciio/_hierarchical.py
+++ b/rsciio/_hierarchical.py
@@ -514,21 +514,22 @@ class HierarchicalReader:
                 dictionary[key] = value
         if not isinstance(group, self.Dataset):
             for key in group.keys():
-                if key.startswith("_sig_"):
+                if key == "ragged_shapes":
+                    # array used to parse ragged array, need to skip it
+                    # otherwise, it will wrongly read kwargs when reading
+                    # variable length markers as they uses ragged arrays
+                    pass
+                elif key.startswith("_sig_"):
                     dictionary[key] = self.group2signaldict(group[key])
                 elif isinstance(group[key], self.Dataset):
                     dat = group[key]
                     kn = key
                     if key.startswith("_list_"):
-                        if h5py.check_string_dtype(dat.dtype) and hasattr(dat, "asstr"):
-                            # h5py 3.0 and newer
-                            # https://docs.h5py.org/en/3.0.0/strings.html
-                            dat = dat.asstr()[:]
-                        ans = np.array(dat)
+                        ans = self._parse_iterable(dat)
                         ans = ans.tolist()
                         kn = key[6:]
                     elif key.startswith("_tuple_"):
-                        ans = np.array(dat)
+                        ans = self._parse_iterable(dat)
                         ans = tuple(ans.tolist())
                         kn = key[7:]
                     elif dat.dtype.char == "S":
@@ -573,6 +574,14 @@ class HierarchicalReader:
                     self._group2dict(group[key], dictionary[key], lazy=lazy)
 
         return dictionary
+
+    @staticmethod
+    def _parse_iterable(data):
+        if h5py.check_string_dtype(data.dtype) and hasattr(data, "asstr"):
+            # h5py 3.0 and newer
+            # https://docs.h5py.org/en/3.0.0/strings.html
+            data = data.asstr()[:]
+        return np.array(data)
 
 
 class HierarchicalWriter:
--- a/rsciio/_hierarchical.py
+++ b/rsciio/_hierarchical.py
@@ -29,7 +29,7 @@ import numpy as np
 from rsciio.utils.tools import ensure_unicode
 
 
-version = "3.2"
+version = "3.3"
 
 default_version = Version(version)
 
@@ -225,6 +225,22 @@ class HierarchicalReader:
 
         return exp_dict_list
 
+    @staticmethod
+    def _read_array(group, dataset_key):
+        data = group[dataset_key]
+        key = f"_ragged_shapes_{dataset_key}"
+        if "ragged_shapes" in group:
+            # For file saved with rosettaSciIO <= 0.1
+            # rename from `ragged_shapes` to `_ragged_shapes_{key}` in v3.3
+            key = "ragged_shapes"
+        if key in group:
+            ragged_shape = group[key]
+            new_data = np.empty(shape=data.shape, dtype=object)
+            for i in np.ndindex(data.shape):
+                new_data[i] = np.reshape(data[i], ragged_shape[i])
+            data = new_data
+        return data
+
     def group2signaldict(self, group, lazy=False):
         """
         Reads a h5py/zarr group and returns a signal dictionary.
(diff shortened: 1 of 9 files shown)


jlaehne (approved):

ericpre:
@CSSFrancis, do you want to review this PR?

CSSFrancis:
@ericpre I can do that now!  Sorry, I've been trying to finish writing some papers the last couple of days so I haven't been the best at responding :)

CSSFrancis:
@ericpre I looked over this briefly and it looks like a really good improvement overall!  I've been meaning to go back to the idea of saving and loading ragged arrays as I think that it is a bit problematic.

One thing that is a bit concerning is that loading ragged arrays into memory is __very slow__ for the `zspy` format. It's not so much an issue when saving the ragged arrays but the loading time appears to be much slower than the time to save the data. It also increases exponentially with the number of markers to make things extra bad :).


For example:

```python 

import time 
import matplotlib.pyplot as plt
import hyperspy.api as hs
import numpy as np

save_times_z = []
load_times_z =[]
save_times_h = []
load_times_h =[]

num_pos = [100, 500, 1000, 2000, 4000, 10000]
load_times =[]
for i in num_pos:
    test = np.empty((i), dtype=object)
    for j in np.ndindex(test.shape):
        test[j]=np.array([[1,1],[2,2]])
        
    s = hs.signals.BaseSignal(test)
    
    tic = time.time()
    s.save("data.zspy", overwrite=True)
    toc =time.time()
    save_times_z.append(toc-tic)
    
    tic = time.time()
    hs.load("data.zspy")
    toc =time.time()
    load_times_z.append(toc-tic)
    
    tic = time.time()
    s.save("data.hspy", overwrite=True)
    toc =time.time()
    save_times_h.append(toc-tic)
    
    tic = time.time()
    hs.load("data.hspy")
    toc =time.time()
    load_times_h.append(toc-tic)

plt.plot(num_pos, load_times_z, label="loading time (zspy)")
plt.plot(num_pos, save_times_z, label="saving time (zspy)")
plt.plot(num_pos, load_times_h, label="loading time (hspy)")
plt.plot(num_pos, save_times_h, label="saving time (hspy")
plt.xlabel("number of positions")
plt.ylabel("time in sec")
plt.legend()
```

image

I realize that this is probably something up stream in `zarr` and not related to this PR but I think that it will cause problems with saving markers like this.

rsciio/_hierarchical.py
```diff
@@ -514,21 +527,22 @@ def _group2dict(self, group, dictionary=None, lazy=False):
                 dictionary[key] = value
         if not isinstance(group, self.Dataset):
             for key in group.keys():
-                if key.startswith("_sig_"):
+                if key.startswith("_ragged_shapes_"):
+                    # array used to parse ragged array, need to skip it
```
CSSFrancis, line 531: This is related to https://github.com/zarr-developers/zarr-specs/issues/62  Support for n-d ragged arrays might be coming soon(ish)

Maybe add the reference there?

rsciio/_hierarchical.py
```diff
@@ -225,6 +225,22 @@ def read(self, lazy):
 
         return exp_dict_list
 
+    @staticmethod
+    def _read_array(group, dataset_key):
+        data = group[dataset_key]
```
CSSFrancis, line 230: Something about this is slow for a `zarr` dataset (and was before this change)

CSSFrancis (commented):
This looks like a good change!  It also cleans up some stuff.  The only thing I am worried about is the slow loading for ragged .zarr arrays. 

This has the problem of potentially "bricking" the loading for a 4-D STEM dataset if you save a large ragged array alongside it and then can't access the data because the loading time for the ragged array is very slow.

CSSFrancis:
@ericpre Let me know if you have any thoughts on why this might be otherwise I can try to look into this more to figure out some way to save/load efficiently.

ericpre:
> @ericpre Let me know if you have any thoughts on why this might be otherwise I can try to look into this more to figure out some way to save/load efficiently.

I don't think that there is much to be done here, as you said, it must come from zarr/numcodecs.

rsciio/_hierarchical.py
ericpre: I have put the link in the `_read_array` function.

CSSFrancis:
So it seems like the VLenArray numcodec isn't the problem as:

```python
def benchmark_codec(codec, a):
    print(codec)
    print('encode')
    %timeit codec.encode(a)
    enc = codec.encode(a)
    print('decode')
    %timeit codec.decode(enc)
    print('size         : {:,}'.format(len(enc)))

np.random.seed(42)
data4 = np.array([np.random.random(size=np.random.randint(0, 20)).astype(np.float64)
                  for i in range(200000)], dtype=object)
data4.shape
benchmark_codec(vlen_arr_codec, data4)
```

This seems to work just fine and scales nicely. The issue seems to instead be that each array is being treated as a separate chunk and the `_decode_chunk` function is being called for every position. 

If we set the number of chunks here to 1:

https://github.com/hyperspy/rosettasciio/blob/e4e71adc89a5fb15dbfc623af50ed430dffe5622/rsciio/zspy/_api.py#L88-L97

Then things work a little better :)
image

CSSFrancis:
@ericpre What do you think.

I would vote that we force any numpy array with `dtype=object` into one chunk for compression/loading.

For a `dask` array we can leave it's chunking scheme and assume that the person saving/loading the data has some idea of what they are doing.

ericpre:
Thank you @CSSFrancis for looking at this. I tried using one chunks in `_get_object_dset` but reading is still slow, maybe you changed somewhere else? 
Do you want to sort out this issue in a separate PR? The issue is not introduced in this PR, which is about getting the test suite to work with the markers update!

CSSFrancis:
> Thank you @CSSFrancis for looking at this. I tried using one chunks in `_get_object_dset` but reading is still slow, maybe you changed somewhere else?
> Do you want to sort out this issue in a separate PR? The issue is not introduced in this PR, which is about getting the test suite to work with the markers update!

Yep I can do that!

Do you want to merge this then and then I will open a new PR.

CSSFrancis (approved):

rsciio/_hierarchical.py
CSSFrancis: Move to new PR.

Issue #168, opened 2023-10-05
Ragged Array Reading is Slow(er than it should be)


CSSFrancis:
#### Describe the bug
There appears to be a bug somewhere in `zarr` (and I would imagine `h5py` based on the fact that the loading speed is similar) with regards to loading ragged arrays.

#### To Reproduce
```python
data = np.array([np.random.randint(0, 100, size=np.random.randint(0, 20)).astype(np.float64)
                  for i in range(i)], dtype=object)
s = hs.signals.BaseSignal(data)
s.save("data.zspy", overwrite=True)
%prun hs.load("data.zspy")
```

#### Expected behavior

Ideally some chunk would be compressed once and then loaded once on each save/load cycle.  What is actually happening is different for `.zspy` and `.hspy`.

For `.hspy` each index in the ragged array is compressed individually and then uncompressed individually.  This isn't efficient at all but isn't the worst case scenario. 

For `.zspy` multiple indexes in the ragged array are compressed at one time but it seems like only 1 index is being uncompressed at a time.  The result is that as you increase the number of indexes compressed (n) the time to uncompress is multiplied by (n).

#### Python environement:
 - RosettaSciIO version: 0.0.2dec
 - Python version: 3.9

#### Additional context
See #164 for more context

ericpre:
Done in https://github.com/hyperspy/rosettasciio/pull/169.

PR #169, opened 2023-10-05
Faster Ragged Reading of Markers

CSSFrancis:
### Description of the change
This fixes (some) of the bugs brought up in #168  #164 

### Progress of the PR
- [x] ready for review.

This should be much faster than previous. (For large arrays this makes ragged arrays __actually__ usable)  I wanted to start adding some benchmarks but I'm not sure the best way to do that? Jupyter notebooks don't seem like the best way to do that, maybe something similar to examples?

```python
import time 
import matplotlib.pyplot as plt
import hyperspy.api as hs
import numpy as np

save_times_z = []
load_times_z =[]
save_times_h = []
load_times_h =[]

num_pos = [100, 500, 1000, 2000, 4000, 10000, 40000, 100000,1000000]
load_times =[]
for i in num_pos:
    np.random.seed(42)
    data = np.array([np.random.random(size=np.random.randint(0, 20)).astype(np.float64)
                  for i in range(i)], dtype=object)
        
    s = hs.signals.BaseSignal(data)
    
    tic = time.time()
    s.save("data.zspy", overwrite=True)
    toc =time.time()
    save_times_z.append(toc-tic)
    
    tic = time.time()
    hs.load("data.zspy")
    toc =time.time()
    load_times_z.append(toc-tic)
    
    tic = time.time()
    s.save("data.hspy", overwrite=True)
    toc =time.time()
    save_times_h.append(toc-tic)
    
    tic = time.time()
    hs.load("data.hspy")
    toc =time.time()
    load_times_h.append(toc-tic)

plt.plot(num_pos, load_times_z, label="loading time (zspy)")
plt.plot(num_pos, save_times_z, label="saving time (zspy)")
plt.plot(num_pos, load_times_h, label="loading time (hspy)")
plt.plot(num_pos, save_times_h, label="saving time (hspy")
plt.xlabel("number of positions")
plt.ylabel("time in sec")
plt.legend()
```


rsciio/zspy/_api.py
```diff
@@ -88,7 +90,15 @@ def __init__(self, file, signal, expg, **kwargs):
...
-        """Creates a Zarr Array object for saving ragged data"""
+        """Creates a Zarr Array object for saving ragged data
+
+        Forces the number of chunks span the array if not a dask array as
+        calculating the chunks for a ragged array is not supported.
+        """
+        if isinstance(data, da.Array):
+            chunks = chunks
+        else:
+            chunks = np.prod(data.shape)
```
ericpre: To remove the codeQL failure:
```suggestion
        if not isinstance(data, da.Array):
            chunks = np.prod(data.shape)
```

rsciio/_hierarchical.py
```diff
@@ -239,7 +239,11 @@ def _read_array(group, dataset_key):
             key = "ragged_shapes"
         if key in group:
             ragged_shape = group[key]
+            ragged_shape = np.array(
+                ragged_shape
+            )  # cast to numpy array to stop multiple reads
```
ericpre, line 244: It is not obvious what is meant with the comment (at least to me!) Can you elaborate a bit more please?

rsciio/zspy/_api.py
```diff
@@ -88,7 +90,15 @@ def __init__(self, file, signal, expg, **kwargs):
 
     @staticmethod
     def _get_object_dset(group, data, key, chunks, **kwds):
-        """Creates a Zarr Array object for saving ragged data"""
+        """Creates a Zarr Array object for saving ragged data
+
+        Forces the number of chunks span the array if not a dask array as
+        calculating the chunks for a ragged array is not supported.
```
ericpre, line 96: Link to https://github.com/hyperspy/rosettasciio/issues/168?

ericpre (commented):
This looks good. When I tried, I used `len` instead of `np.prod` and it seems that it was the reason why it was still slow!

I am not sure about how useful the benchmark file would be, because we may well forget about it if there isn't something that run regularly and check how long it takes.
Maybe a test with a upper limit on the execution time would be more useful?

rsciio/_hierarchical.py
CSSFrancis: Done!

rsciio/zspy/_api.py
CSSFrancis: Yea I'm not sure why I wrote that like that... Maybe coding too early in the morning.

CSSFrancis:
@ericpre I'll just remove the Jupyter notebook for now. I think the information is in this PR and 168 to recreate the tests so that should be good. 

Maybe something to come back to in time.  I think that some sort of benchmarking for the different file loaders would be helpful as it would help to identify which ones are faster/slower and help to identify if certain file readers could be made faster.

ericpre (approved):

PR #173, opened 2023-10-16
bugfix: included the "Out" case for the EELS aperture size

EnigmaticEureka:
The previous code did not account for Acquisition_instrument.TEM.Detector.EELS.aperture_size being "Out" therefore running into a problem upon trying to convert this string into a float.

It has now been altered such that if "mm" is not in aperture, the string will be added to the meta data unchanged (as "Out").

### DIFF
--- a/rsciio/pantarhei/_api.py
+++ b/rsciio/pantarhei/_api.py
@@ -349,9 +349,13 @@ def _metadata_converter_in(meta_data, axes, filename):
         if "mm" in aperture:
             aperture = aperture.split("mm")[0]
             aperture = aperture.rstrip()
-        mapped.set_item(
-            "Acquisition_instrument.TEM.Detector.EELS.aperture_size", float(aperture)
-        )
+            mapped.set_item(
+                "Acquisition_instrument.TEM.Detector.EELS.aperture_size", float(aperture)
+            )
+        else:
+            mapped.set_item(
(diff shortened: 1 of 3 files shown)


rsciio/pantarhei/_api.py
```diff
@@ -349,9 +349,13 @@ def _metadata_converter_in(meta_data, axes, filename):
...
             aperture = aperture.split("mm")[0]
             aperture = aperture.rstrip()
-        mapped.set_item(
-            "Acquisition_instrument.TEM.Detector.EELS.aperture_size", float(aperture)
-        )
+            mapped.set_item(
+                "Acquisition_instrument.TEM.Detector.EELS.aperture_size", float(aperture)
+            )
+        else:
+            mapped.set_item(
```
ericpre, line 356: It would be good to add a test to cover this line, for example, write a test which create a file with such aperture and read it again.
If you are not familiar with writing tests, you can have a look at https://hyperspy.readthedocs.io/en/latest/dev_guide/testing.html. The test should go in https://github.com/hyperspy/rosettasciio/blob/main/rsciio/tests/test_pantarhei.py. Of course, you can use other tests as source of inspiration!

ericpre (commented):
Thanks @EnigmaticEureka, this is a good start! :)
I left a comment regarding adding a test and it also needs a changelog entry - see https://github.com/hyperspy/rosettasciio/blob/main/upcoming_changes/README.rst for guidance.

PR #174, opened 2023-10-16
Mib reader

ericpre:
This PR a mib reader with the code adapted from https://github.com/matkraj/read_mib/ and I took the example files from that repository, thanks @matkraj! And also some other examples from https://github.com/pyxem/pyxem/pull/732#issuecomment-836527568, thanks @magnunor!

### Progress of the PR
- [x] Add mib reader for single and quad chips,
- [x] add support for reading interrupted acquisition,
- [x] add tests,
- [x] ready for review.

Unlike the existing reader in pyxem, the implementation of this PR doesn't support the two following features:
- add gap quadrant for camera with four quadrant (to consider for pyxem with interpolation too?)

This shouldn't be that difficult to add these but I don't know if this is still necessary - I am hoping that recent versions of acquisition software can generate files that doesn't need these. Reading file without using these features will be faster, so better to avoid these if possible!

For follow-up PR:
- add support for reading ROI
- infer navigation shape from flyback - reading exposure time is implemented but test files are needed.
- add support for reading RAW data


CSSFrancis:
@ericpre Thanks for doing this!  Let me know when you want someone to review it and I can give it a shot.

Would it be worth adding support for distributed reading of the `mib` files similar to what I have been doing with the .mrc files in #162? 

I might do a follow up PR adding that if it seems worthwhile.

ericpre:
> @ericpre Thanks for doing this! Let me know when you want someone to review it and I can give it a shot.

It seems that the coverage still need to be increased! 😅 

> Would it be worth adding support for distributed reading of the `mib` files similar to what I have been doing with the .mrc files in #162?
> 
> I might do a follow up PR adding that if it seems worthwhile.

Yes, definitely, it did cross my mind! 😃

ericpre:
pre-commit.ci autofix

ericpre:
@CSSFrancis, this is ready to review.

magnunor:
I can try this on a couple of mib-files I have.

Does this include support for the "raw" mode? In this mode, it dumps the data from the detector directly to the file, without correcting for the wrong order of the data.

ericpre:
> I can try this on a couple of mib-files I have.
> 
> Does this include support for the "raw" mode? In this mode, it dumps the data from the detector directly to the file, without correcting for the wrong order of the data.

No, it raises an error. Would it that be useful? I wasn't thinking of supporting and but of course anyone can do it in a follow up PR!

magnunor:
> No, it raises an error. Would it that be useful? I wasn't thinking of supporting and but of course anyone can do it in a follow up PR!

It would be nice, as running in raw mode enables for a higher data rate over longer times. (Or at least last time I tried it). But this can be left for a future pull request.

I tested the mib-reader in this pull request on some mib-file, and there seems to be some issues somewhere:

pyxem_load_mib_vs_this_pull_request

The "mirroring" can be ignored, but there is something wrong with the data values. Probably related to some dtype issue.

I don't have a mib-dataset at hand which I can share right now, but I'll see if I can dig one up.

ericpre:
> The "mirroring" can be ignored, but there is something wrong with the data values. Probably related to some dtype issue.

What is the dtype of the data for both? Can you give what `s.data.dtype` is?

ericpre:
Thanks @magnunor: it was the endianness which was being lost!

Regarding orientation of the signal space, this is the same as https://github.com/matkraj/read_mib/blob/main/mib_examples/quad.png.

matkraj:
The Raw data should be only useful for 1-bit imaging. I thought about some
efficient algorithms to do the conversion and I think the best is to load
the file as a 64bit (making frame length a 1024 pixels double) and then run
endian swapping (byteswap). After this the data can be extracted with
unpackbits and reshaping this to 256x256 array.

matkraj:
To not forget the quad chips, the data comes out in specific chip order and
this will also need to be reshaped. I can’t recall what it is at the moment
but using one of the example files from my repo should be straightforward.
Hope these hints can help towards the full support of the format.
Thanks for looking into this area, Matus

ericpre:
Thank you @matkraj, this sounds useful, I will add it to the list for follow up PR!

CSSFrancis (commented):
@ericpre This looks like a good addition.  @magnunor probably did more thorough testing than I can as he has access to some .mib files :). The only thing that I'm a little hesitant about is things like reshaping and flipping data. 

Memory mapping is fastest when it is reading data in a line.  For the case of 4-D STEM data the fastest way to load the data is first by the diffraction pattern and second along the row of acquisition. 

For example:

```python
test = np.memmap("test.mem", dtype="float32", mode="c", shape=(40,40,40,40))
%%timeit
np.array(np.flip(test[1,:,:,:], axis=-2))

#20.7 µs ± 330 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
%%timeit
np.flip(np.array(test[1,:,:,:]),axis=-2)
# 10.5 µs ± 42.2 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```
My understanding is that numpy handles these kinds of operations on `memmap` objects by jumping around on the disk.  This is fine in most cases but isn't terribly efficient as it depends on how fast your disk reading is rather than how fast your memory is.  

The better thing (I think) is to load the array using a fast memory mapping implementation and then use dask to flip etc. if lazy and use numpy to flip etc if not.

magnunor:
> Thanks @magnunor: it was the endianness which was being lost!

That fixed the strange values :)

I can grab some small quadchip datasets, with different chips activated. To see if they also work. Maybe today.

rsciio/quantumdetector/_api.py
```diff
@@ -269,18 +275,26 @@ def load_mib_data(
...
+
     else:  # pragma: no cover
-        raise ValueError("`path` must be a str or a buffer.")
+        raise TypeError("`path` must be a str or a buffer.")
+
+    headers = data["header"]
+    data = data["data"]
+    if not return_mmap:
+        if lazy:
+            data = da.from_array(data)
```
CSSFrancis, line 286: Add in chunks parameter here?

ericpre: Yes, quite often, we read from files using chunking and there is limited advantage in specifying the chunking but in this case, it make sense!

We should document it and make more consistent across file readers.

ericpre (commented):
Thank you @CSSFrancis, these are very good point! It should be significantly better now!

Regarding handling of the `numpy.memmap`, will you be revisiting some of it as part of #162 or another PR?

magnunor:
Testing lazy vs non-lazy, via HyperSpy `load` there are some differences:

Non-lazy:

```python
s = hs.load("my8600mbdataset.mib) # this uses about 8.6 GB of memory
```

Lazy:

```python
s = hs.load("my8600mbdataset.mib, lazy=True)
s.compute() # This uses about 16.5 GB
```

-----------

Trying to save one of the datasets from my previous comment:

```python
s = hs.load("default_12_bit_chip_all_4x2_frames.mib")
s.save("test.hspy")
```
Gives the error:

```python
The writer could not write the following information in the file: dtype : >u2
Traceback (most recent call last):
  File "rosettasciio/rsciio/_hierarchical.py", line 820, in dict2group
    group.attrs[key] = value
    ~~~~~~~~~~~^^^^^
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/_hl/attrs.py", line 104, in __setitem__
    self.create(name, data=value)
  File h5py/_hl/attrs.py", line 182, in create
    htype = h5t.py_create(original_dtype, logical=True)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "h5py/h5t.pyx", line 1658, in h5py.h5t.py_create
  File "h5py/h5t.pyx", line 1682, in h5py.h5t.py_create
  File "h5py/h5t.pyx", line 1742, in h5py.h5t.py_create
TypeError: Object dtype dtype('O') has no native HDF5 equivalent

ericpre:
Thinking more of the "add gap quadrant" feature, I lean towards doing this in pyxem and at the same time, try to add interpolation of intensities? This would be seen as the data correction and therefore shouldn't be in rosettasciio, which is only concerned with reading data.

CSSFrancis:
> > Testing lazy vs non-lazy, via HyperSpy `load` there are some differences:
> > Non-lazy:
> > ```
> > s = hs.load("my8600mbdataset.mib) # this uses about 8.6 GB of memory
> > ```
> > 
> > 
> > Lazy:
> > ```
> > s = hs.load("my8600mbdataset.mib, lazy=True)
> > s.compute() # This uses about 16.5 GB
> > ```
> 
> I have seen this before and I don't know if there is much to be done here: for the non-lazy case, I get a peak memory of ~14GB when reading a ~8GB file. For lazy case, it seems to behave better in term of speed memory usage than the `pyxem.load_mib`. @CSSFrancis, do you want to chime in on this point?

I'm not exactly sure what is going on there either.  My best guess is that the reshape function that `dask` is using might end up rechunking the data which would cause the memory spike. That or maybe  sometimes `dask` doesn't clean up after itself if it doesn't need to and isn't feeling any memory pressure.

ericpre:
> I'm not exactly sure what is going on there either. My best guess is that the reshape function that `dask` is using might end up rechunking the data which would cause the memory spike. That or maybe sometimes `dask` doesn't clean up after itself if it doesn't need to and isn't feeling any memory pressure.

I have tried without reshaping and it is the same. I have the impression that there is some memory not being release because I can load the dataset non-lazily even if it seems that there is no enough memory - if this was the case, I would get a memory error. This is on windows, which usually is worse than on other system but the behaviour seems to be consistent with @magnunor's description above.

magnunor:
> Thinking more of the "add gap quadrant" feature, I lean towards doing this in pyxem and at the same time, try to add interpolation of intensities? This would be seen as the data correction and therefore shouldn't be in rosettasciio, which is only concerned with reading data.

Yeah, I agree. I feel that `rosettasciio` should not be doing those kinds of corrections, and those belong in libraries such as pyxem.

ericpre:
@magnunor, @CSSFrancis, if you don't have any additional comments on this PR, I will merge it in the coming days.

CSSFrancis (approved):

CSSFrancis:
@ericpre lgtm!

magnunor:
@ericpre looks good to me as well!

PR #175, opened 2023-10-19
Pin pillow<10.1.0 until imageio is fixed

ericpre:
See https://github.com/imageio/imageio/issues/1044.


ericpre:
Merging as this is trivial and in the interest of fixing the test suite!

PR #180, opened 2023-11-02
Make numba and h5py optional dependencies

ericpre:
In addition to reducing the number of required dependencies, this also has the advantage of testing the lastest python version before numba supports it but also to run rosettasciio in JupyterLite.

### Progress of the PR
- [x] Make numba and h5py optional dependencies,
- [x] test against python 3.12,
- [ ] add tests,
- [ ] ready for review.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,6 +18,7 @@ classifiers = [
   "Programming Language :: Python :: 3.9",
   "Programming Language :: Python :: 3.10",
   "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
   "Operating System :: Microsoft :: Windows",
   "Operating System :: POSIX",
   "Operating System :: Unix",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,17 +26,14 @@ classifiers = [
 dependencies = [
   "dask[array]>=2021.3.1",
   "python-dateutil",
-  "h5py>=2.3",
   "imageio>=2.16",
   # Remove when https://github.com/imageio/imageio/issues/1044 is fixed
   "pillow<10.1.0",
-  "numba>=0.52",
   "numpy>=1.20.0",
   "pint>=0.8",
   "python-box>=6,<7",
   "pyyaml",
   "scipy>=1.5.0",
-  "sparse",
 ]
 dynamic = ["version"]
 
@@ -50,6 +47,7 @@ file = "COPYING.txt"
 
 [project.optional-dependencies]
 blockfile = ["scikit-image>=0.18"]
+hdf5 = ["h5py>=2.3"]
 mrcz = ["blosc>=1.5", "mrcz>=0.3.6"]
 scalebar_export = ["matplotlib-scalebar", "matplotlib>=3.5"]
 tiff = ["tifffile>=2020.2.16", "imagecodecs>=2020.1.31"]
@@ -74,12 +72,15 @@ docs = [
   "towncrier"
 ]
 all = [
+  "numba>=0.52",
   "rosettasciio[blockfile]",
+  "rosettasciio[hdf5]",
   "rosettasciio[mrcz]",
   "rosettasciio[scalebar_export]",
   "rosettasciio[tiff]",
   "rosettasciio[usid]",
   "rosettasciio[zspy]",
+  "sparse",
 ]
 dev = [
   "black",
--- a/rsciio/utils/tools.py
+++ b/rsciio/utils/tools.py
@@ -516,3 +516,21 @@ def get_file_handle(data, warn=True):
                     "an hdf5 file."
                 )
     return None
+
+
+def jit_ifnumba(*args, **kwargs):
+    try:
+        import numba
+
+        if "nopython" not in kwargs:
+            kwargs["nopython"] = True
+        return numba.jit(*args, **kwargs)
+    except ImportError:
+
+        def wrap1(func):
+            def wrap2(*args2, **kwargs2):
+                return func(*args2, **kwargs2)
+
+            return wrap2
+
+        return wrap1
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -28,7 +28,7 @@ jobs:
             PYTHON_VERSION: '3.9'
             LABEL: '-minimum'
           - os: ubuntu
-            PYTHON_VERSION: '3.9'
+            PYTHON_VERSION: '3.12'
             LABEL: '-minimum-wo-hyperspy'
           - os: ubuntu
             PYTHON_VERSION: '3.9'
--- a/rsciio/jeol/_api.py
+++ b/rsciio/jeol/_api.py
@@ -22,10 +22,9 @@ from datetime import datetime, timedelta
 import logging
 
 import numpy as np
-import numba
 
 from rsciio._docstrings import FILENAME_DOC, LAZY_DOC, RETURNS_DOC
-
+from rsciio.utils.tools import jit_ifnumba
 
 _logger = logging.getLogger(__name__)
 
@@ -984,7 +983,7 @@ def _readcube(
     )
 
 
-@numba.njit(cache=True)
+@jit_ifnumba(cache=True)
 def _readframe_dense(
     rawdata,
     countup,
@@ -1097,7 +1096,7 @@ def _readframe_dense(
     return count, 0, has_em_image, valid, previous_y // height_norm
 
 
-@numba.njit(cache=True)
+@jit_ifnumba(cache=True)
 def _readframe_lazy(
     rawdata,
     _1,
@@ -1198,7 +1197,7 @@ def _readframe_lazy(
     return count, data, has_em_image, valid, previous_y // height_norm
 
 
-@numba.njit(cache=True)
+@jit_ifnumba(cache=True)
 def _skip_frame(rawdata):  # pragma: no cover
     count = 0
     previous_y = 0
--- a/rsciio/tests/test_emd.py
+++ b/rsciio/tests/test_emd.py
@@ -24,17 +24,18 @@
 import os
 from pathlib import Path
 
+import pytest
+
+hs = pytest.importorskip("hyperspy.api", reason="hyperspy not installed")
+
 import dask.array as da
 from datetime import datetime
 from dateutil import tz
 import gc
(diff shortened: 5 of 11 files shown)


CSSFrancis:
@ericpre Have you installed the wheel through `micropip` I can't seem to build a pure python wheel (rosettasciio-0.2.0-py3-none-any.whl) as needed for micropip to install some package.  

It would be good to add to the `release.yml` to build a [pure python wheel](https://pyodide.org/en/stable/development/building-and-testing-packages.html#building-binary-packages-for-pyodide)

There might be another way around this?

CSSFrancis:
Also just wanted to say I'm __super__ excited about this :)

CSSFrancis:
@ericpre I tried to get this working in Jupyter lite and I wonder if you have had any success?


Just for note of how I got this to work because I kind of struggled a bit previously. 


1. install pipx (brew install pipx... etc. )
2. run %DISABLE_C_EXTENTIONS=1 pipx run build --wheel
3. start a Jupyterlite instance (https://jupyter.org/try-jupyter/lab/?path=notebooks%2FIntro.ipynb)
4. Move the `rosettasciio-0.2.dev0-py3-none-any.whl` file to the Jupyter lite instance using the upload a file function
5. Run:
```python
import micropip
await micropip.install("emfs:rosettasciio-0.2.dev0-py3-none-any.whl")
```

if you then try:
```python
import rsciio
```

The following error is raised.

```
OSError                                   Traceback (most recent call last)
Cell In[10], line 1
----> 1 import rsciio

File /lib/python3.11/site-packages/rsciio/__init__.py:56
     53 import os.path
     55 try:
---> 56     p = subprocess.Popen(
     57         ["git", "log", "-1", '--format="%h %aI"'],
     58         stdout=subprocess.PIPE,
     59         stderr=subprocess.PIPE,
     60         cwd=os.path.dirname(__file__),
     61     )
     62 except FileNotFoundError:
     63     pass

File /lib/python311.zip/subprocess.py:818, in Popen.__init__(self, args, bufsize, executable, stdin, stdout, stderr, preexec_fn, close_fds, shell, cwd, env, universal_newlines, startupinfo, creationflags, restore_signals, start_new_session, pass_fds, user, group, extra_groups, encoding, errors, text, umask, pipesize, process_group)
    816 """Create new Popen instance."""
    817 if not _can_fork_exec:
--> 818     raise OSError(
    819         errno.ENOTSUP, f"{sys.platform} does not support processes."
    820     )
    822 _cleanup()
    823 # Held while anything is calling waitpid before returncode has been
    824 # updated to prevent clobbering returncode if wait() or poll() are
    825 # called from multiple threads at once.  After acquiring the lock,
    826 # code must re-check self.returncode to see if another thread just
    827 # finished a waitpid() call.

OSError: [Errno 138] emscripten does not support processes.
```

CSSFrancis:
@ericpre seems like we should just  be able to skip the subprocesses and things should work. It might be good to think about adding a CI that tests if the package installs via micropip if that is something we are interested in.

ericpre:
Yes, maybe something that run daily or weekly to check that this keep working would be good.
I haven't tried to run in jupyterlite, can you try with the latest commit?

If this works fine, I will document how the version is determined.

CSSFrancis:
@ericpre Did something get changed there? The version is not 0.1.0dev and it's packaging the tests in the wheel.

CSSFrancis:
For testing maybe we consider something like sympy.  They have a pretty complete pyodide check.


[javescript file for creating tests](https://github.com/sympy/sympy/blob/master/bin/test_pyodide.mjs)

ericpre:
For the packaging issue, there is the same issue (tests file being packaged) with v0.1, so it must be related to a change in setuptools and co.

I added the `MANIFEST.in` to remove these and it seems to be fine now - at least for now...

CSSFrancis:
> For the packaging issue, there is the same issue (tests file being packaged) with v0.1, so it must be related to a change in setuptools and co.
> 
> I added the `MANIFEST.in` to remove these and it seems to be fine now - at least for now...

Hmm I still have it including the test files so the wheel is ~80 mb in size...

CSSFrancis:
@ericpre Sorry I'm not very much help here.  I'm no vacation (until tomorrow) so I haven't had much time to look over this in detail :)

ericpre:
No worries, on CI the packaging doesn't include the test files, so we should be good!

ericpre:
@CSSFrancis: you will like this: https://ericpre.github.io/jupyterlite-hyperspy!

CONTRIBUTING.rst
```diff
@@ -167,6 +167,29 @@ GitHub CI workflow is set to fail when the doc build raises warnings.
...
+- Version at build time: the version is defined from the tag or the "distance from the tag".
+- Version at runtime: use the version of the package (``sdist`` or ``wheel``), which would have been
+  defined at build time. At runtime, the version is obtained using importlib.metadata as follow:
+
+  .. code-block:: python
+  
+    from importlib.metadata import version
+    __version__ = version("rosettasciio")
+
+- Version at runtime for editable installation: the version is defined from the tag or"the distance from the tag".
```
CSSFrancis, line 184: ** missing a space
```suggestion
- Version at runtime for editable installation: the version is defined from the tag or "the distance from the tag".
```

CSSFrancis (approved):
Just some small things like spaces etc.  Otherwise this looks good!

docs/user_guide/install.rst
```diff
@@ -0,0 +1,101 @@
...
+installed with `pip <https://pip.pypa.io/en/stable>`__.
+To install, run the following::
+
+    pip install rosettasciio
+
+To update rosettasciio to the latest release::
+
+    pip install --upgrade rosettasciio
+
+To install a specific version of kikuchipy (say version 0.1)::
```
francisco-dlp, line 27: kikuchipy -> hyperspy?

docs/user_guide/install.rst
```diff
@@ -0,0 +1,101 @@
...
+:doc:`conda environment <conda:user-guide/tasks/manage-environments>` with the
+`Miniforge distribution <https://github.com/conda-forge/miniforge>`_.
+To create an environment and activate it, run the following::
+
+   conda create --name rsciio python=3.11
+   conda activate rsciio
+
+To install::
+
+    conda install rosettasciio
```
francisco-dlp, line 72: `conda` doesn't support `extras`. It would be good to clarify this point, and explain what is installed when using `conda`.

ericpre: Yes, we should make a `rosettasciio-base` (no optional dependencies) and a `rosettasciio` (all optional dependencies) on `conda-forge`.

rsciio/utils/tools.py
```diff
@@ -516,3 +516,21 @@ def get_file_handle(data, warn=True):
...
+
+
+def jit_ifnumba(*args, **kwargs):
+    try:
+        import numba
+
+        if "nopython" not in kwargs:
+            kwargs["nopython"] = True
+        return numba.jit(*args, **kwargs)
+    except ImportError:
```
francisco-dlp, line 528: We could add a logger warning here to explain that the code is not running with optimal performance because number is not installed.

francisco-dlp (changes requested):
Other than the comments above, I think that it would be good to clarify in the documentation the precise policy we use to decide which features to put into `extras`. In particular, I assume that the main aim of this PR is to enable running `RosettaSciIO` in JupyterLite with the default installation. I think that this should be stated explicitly in the documentation if that's the case. Once some of the requirements in extra are supported by JupyterLite, will we move them back to the default installation?

pyproject.toml
```diff
@@ -26,17 +27,14 @@ classifiers = [
...
   "imageio>=2.16",
   # Remove when https://github.com/imageio/imageio/issues/1044 is fixed
   "pillow<10.1.0",
-  "numba>=0.52",
   "numpy>=1.20.0",
   "pint>=0.8",
   "python-box>=6,<7",
   "pyyaml",
   "scipy>=1.5.0",
-  "sparse",
```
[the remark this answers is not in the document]
ericpre: `sparse` depends on `numba`!

pyproject.toml
```diff
@@ -62,24 +61,29 @@ tests = [
...
   "pydata-sphinx-theme>=0.13",
+  "setuptools_scm",
   "sphinx",
   "sphinx-favicon",
   "sphinxcontrib-towncrier",
-  "towncrier"
+  "towncrier",
 ]
 all = [
+  "numba>=0.52",
```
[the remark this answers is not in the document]
ericpre: We can add a `speed` extra but `sparse` is for the JEOL and FEI stream reader, which doesn't fit well.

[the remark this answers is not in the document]
ericpre: Went for `speed` and `eds-stream`

docs/user_guide/install.rst
ericpre: https://github.com/conda-forge/rosettasciio-feedstock/pull/3

docs/user_guide/install.rst
```diff
@@ -0,0 +1,119 @@
...
+To install a specific version of RosettaSciIO (say version 0.1)::
+
+    conda install rosettasciio=0.1
+
+.. _install-with-hyperspy-bundle:
+
+With the HyperSpy Bundle
+------------------------
+
+RosettaSciIO is available in the HyperSpy Bundle. See :ref:`hyperspy:hyperspy-bundle` for
```
francisco-dlp, line 105: The HyperSpy Bundle comes with RosettaSciIO and all its extras pre-installed.

francisco-dlp (changes requested):
Looks good. Just some small suggestions regarding the documentation:

* State the precise policy used to decide if a plugin is available by default or goes to `extras`. For example: "depends solely on standard library modules, numpy and dask.". That would also help us decide whether the correct list of `extras` is correct. For example, `tifffile` is a pure Python package that depends solely on `numpy`, however it is in `extras`.
* Explain that `pip` gives maximum installation flexibility and that `pip install rosettasciio` is equivalent to `mamba install rosettasciio-base`, while `mamba install rosettasciio` is equivalent to `pip install rosettasciio[all]`.
* Add a warning to the User Guide entry of each feature/io plugin that require the installation of extra packages.

francisco-dlp (approved):

PR #181, opened 2023-11-08
Release 0.2

ericpre:
Since now `setuptools_scm` is used to set the version at build time from the tag, we don't need to update the version manually and we only need to update the release notes.


ericpre:
Test release workflow:
https://github.com/ericpre/rosettasciio/actions/runs/6518495321/job/17703976497

ericpre:
I will release later today.

ericpre:
https://pypi.org/project/rosettasciio/0.2/
https://anaconda.org/conda-forge/rosettasciio/files?version=0.2

Issue #184, opened 2023-11-19
setuptools_scm missing for some linux install, due to dist-packages location


magnunor:
If `rosettasciio` is installed via PyPi with root in (for example) Ubuntu, it is placed in `/usr/local/lib/python3.10/dist-packages`.

This leads to a problem with importing `rosettasciio`, due to  `rosettasciio` thinking it is in an editable installed (https://github.com/hyperspy/rosettasciio/blob/main/rsciio/__init__.py#L25), excepting the `setuptools_scm` package. However, the `setuptools_scm`  is not installed by default, leading to an error. For example in the Atomap testing system: https://gitlab.com/CSSFrancis/atomap/-/jobs/5567299422#L3567

I guess this is will not occur for very many people, but the fix should be fairly straight forward. For example by changing 

```python
if Path(__file__).parent.parent.name == "site-packages":  # pragma: no cover
```
to

```python
if Path(__file__).parent.parent.name in ["site-packages", "dist-packages"]:  # pragma: no cover
```

# To Reproduce

```
docker run -it ubuntu:latest
apt update
apt install python3-numpy python3-pip
pip install hyperspy==2.0rc0
python3
import rsciio
```

ericpre:
Do you want to make a R?

magnunor:
Should be resolved in https://github.com/hyperspy/rosettasciio/pull/187#event-11123557845

PR #185, opened 2023-11-20
Add dist-packages to check if install is development or not

magnunor:
See https://github.com/hyperspy/rosettasciio/issues/184

### Progress of the PR
- [ ] Change implemented
- [ ] ready for review.

### DIFF
--- a/rsciio/__init__.py
+++ b/rsciio/__init__.py
@@ -21,8 +21,8 @@ import os
 from pathlib import Path
 import yaml
 
-
-if Path(__file__).parent.parent.name == "site-packages":  # pragma: no cover
+folder_names = ["site-packages", "dist-packages"]
+if Path(__file__).parent.parent.name in folder_names:  # pragma: no cover
     # Tested in the "build" workflow on GitHub CI
     from importlib.metadata import version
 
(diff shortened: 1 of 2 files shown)


magnunor:
Some tests related to `exspy`  are failing, which seems to be related to a similiar issue in `exspy`? Ergo that `setuptools-scm` can't find the version number in the build requirements?

`LookupError: setuptools-scm was unable to detect version for /tmp/pip-req-build-8_s83223.`

I got a feeling that way of getting the version is a bit fragile, especially when there are various automated processes which doesn't follow the expected folder/path/... . A possible solution could be to make it return a `None` for the version number if it can't find the git version?

ericpre:
This error is normal, because there are no git repository. The workflow needs to be updated to:
```
pip install git+https://github.com/hyperspy/exspy.git
```
https://packaging.python.org/en/latest/tutorials/installing-packages/#installing-from-vcs

We could fall back to a hardcoded version like `"dev"`? We will not need to keep updating the version when releasing and this would be useful to allow installing from a path without git being available.

magnunor:
Superseded by https://github.com/hyperspy/rosettasciio/pull/187

PR #197, opened 2023-12-12
Release 0.3

ericpre:
https://github.com/hyperspy/rosettasciio/blob/main/releasing_guide.md

PR #199, opened 2023-12-12
Update order of towncrier types

jlaehne:
Update the order of towncrier categories as proposed in  https://github.com/hyperspy/rosettasciio/pull/197#issuecomment-1852678328

### Progress of the PR
- [x] ready for review.


jlaehne:
Possibly "Enhancements" could go up higher in the list.

ericpre:
Agree, "Enhancements" would be good in second position.

ericpre (approved):

PR #200, opened 2023-12-13
Add `POOCH_BASE_URL` to specify the base url used by pooch to download test data

ericpre:
This will fix the `package_and_test.yml` workflow after adding some files in a PR (see for example https://github.com/hyperspy/rosettasciio/pull/162#issuecomment-1830089213) and also the `release.yml` workflow when running from a fork for testing the workflow (it is current working fine for a "real" run because the tag is then available in the `hyperspy/rosettasciio` repository).
I though it would be more complicated to fix and possibly not worth making something complicated but this solution is actually very simple.

### Progress of the PR
- [x] Add `POOCH_BASE_URL` to specify the base url used by pooch to download test data,
- [x] add tests,
- [x] ready for review.

### DIFF
--- a/upcoming_changes/200.maintenance.rst
+++ b/upcoming_changes/200.maintenance.rst
@@ -1,0 +1,1 @@
+Add `POOCH_BASE_URL` to specify the base url used by pooch to download test data. This fixes the failure of the ``package_and_test.yml`` workflow in pull requests where test data are added or updated.
--- a/.github/workflows/package_and_test.yml
+++ b/.github/workflows/package_and_test.yml
@@ -9,3 +9,5 @@ jobs:
     uses: hyperspy/.github/.github/workflows/package_and_test.yml@main
     with:
       module_name: rsciio
+      # github.head_ref is for pull request event while github.ref_name is for push event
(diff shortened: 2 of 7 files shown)


upcoming_changes/200.maintenance.rst
```diff
@@ -0,0 +1 @@
+Add `POOCH_BASE_URL` to specify the base url used by pooch to download test data. This fixes the failure of the ``package_and_test.yml`` workflow in pull requests where test data are added or updated
```
jlaehne, line 1:
```suggestion
Add `POOCH_BASE_URL` to specify the base url used by pooch to download test data. This fixes the failure of the ``package_and_test.yml`` workflow in pull requests where test data are added or updated.
```

jlaehne (approved):

ericpre:
For reference, I had to fiddle with finding the correct github context variable for `push` and `pull_request` event, which is not very from the [github documentation](https://docs.github.com/en/webhooks/webhook-events-and-payloads) but actually event payload can viewed as documented in  https://docs.github.com/en/webhooks/testing-and-troubleshooting-webhooks/viewing-webhook-deliveries.

PR #209, opened 2024-01-02
Fix hamamatsu signal type

jlaehne:
### Description of the change
The hamamatsu reader was trying to set the `signal_type` to the `Signal1D` subclass `Luminescence` instead of the `Signal2D` subclass `TransientSpec` when `lumispy` is available.

### Progress of the PR
- [n/a] add tests,
- [x] ready for review.


rsciio/hamamatsu/_api.py
```diff
@@ -359,7 +359,7 @@ def _map_signal_md(self):
             )
             signal["signal_type"] = ""
         else:
-            signal["signal_type"] = "Luminescence"  # pragma: no cover
+            signal["signal_type"] = "TransientSpec"  # pragma: no cover
```
ericpre, line 362: This is a lumispy implementation details, but wouldn't be better to have "Transcient" as `signal_type` and use the `signal_dimension` to map to the correct class depending if there is an energy axis or not?

jlaehne: The streak camera gives images with one axis being wavelength and one time, so this reader should not come across 1D signals.

ericpre: Sorry, I was not clear: I was suggesting to use `Transient` here in order to use it for both the  `LumiTransient` and `LumiTransientSpectrum` class, which would require changes in lumispy and would fix behaviour reported in https://github.com/LumiSpy/lumispy/issues/204: when changing `signal_dimension`, hyperspy will map to the correct class.

I have just asked myself the same questions as in https://github.com/LumiSpy/lumispy/issues/5 (in particular, does it make sense to have a signal containing dimension with signal type of different nature (energy/wavelength versus time)?

If the discussion on the structure of transient signal can be resolved for good soon, should we wait for the relevant changes in lumispy and update accordingly this PR? If not, we can merge this PR, but I imagine that it will need to be changed when the structure of the transient classes are fixed (currently, I don't think that it is fully compatible with the hyperspy signals paradigm)

jlaehne: OK, my misunderstanding was that `signal_type` is directly linked to the signal class. But a Signal2D can actually have a signal type from a 1D class.

For the moment it would be the quickest fix to directly set `LumiTransientSpectrum` to allow for the energy conversion, but from what you outline, it could also be done downstream by discovering the signal_dimension and then using the correct `LumiTransientSpectrum` class? So currently, there is no added value of using `Transient` here instead of `Luminescence`, but in the long run, there should be dedicated functionalities for the two different signal types combined in one class.

Thus, to resolve this issue it would be a possibility to allow for a list of `signal_type` upstream in hyperspy, where the order determines what axes the signal types correspond to.

upcoming_changes/209.bugfix.rst
```diff
@@ -0,0 +1 @@
+Set ``signal_type`` to ``TransientSpec`` when Hamamatsu ``.img`` files are read and LumiSpy is installed.
```
ericpre, line 1: Does this need updating?
```suggestion
Set ``signal_type`` to ``LumiTransientSpectrum`` when Hamamatsu ``.img`` files are read and LumiSpy is installed.
```

ericpre (approved):

upcoming_changes/209.bugfix.rst
jlaehne: thanks for the catch

PR #211, opened 2024-01-17
Bugfix: Set equal chunking for shapes and dataset

CSSFrancis:
### Description of the change
I noticed this morning that I was having issues loading larger arrays of vectors using the distributed-dask backend.

The problem is that the shape array is not saved with the same chunking scheme so when you try to load the shape array you have to split the memory between all of the different cores.  This is slow, probably partially because dask doesn't know the size of the array which 

### Progress of the PR
- [ ] add tests,
- [ ] ready for review.

```python
pks = s.find_peaks()
pks.save("ragged.zspy")
pks = hs.load("radded.zspy", lazy=True)
pks.compute() 
```

For a 1024x1024 array of peak positions the old saving scheme took 1 minute to load with the new scheme that becomes ~1 second with the new scheme.

### DIFF
--- a/rsciio/_hierarchical.py
+++ b/rsciio/_hierarchical.py
@@ -764,7 +764,7 @@ class HierarchicalWriter:
                     shapes[i] = np.array(data[i].shape)
 
             shape_dset = cls._get_object_dset(
-                group, shapes, f"_ragged_shapes_{key}", shapes.shape, **kwds
+                group, shapes, f"_ragged_shapes_{key}", chunks, **kwds
             )
 
             cls._store_data(
@@ -772,7 +772,7 @@ class HierarchicalWriter:
                 (dset, shape_dset),
                 group,
                 (key, f"_ragged_shapes_{key}"),
-                (chunks, shapes.shape),
+                (chunks, chunks),
                 show_progressbar,
             )
         else:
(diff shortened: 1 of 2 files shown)


ericpre:
Looks good, do you want to add a changelog entry

CSSFrancis:
> Looks good, do you want to add a changelog entry

Yea I'll add a changelog entry, but I might just try it out today to make sure there aren't any other bugs related to this!

rsciio/_hierarchical.py
```diff
@@ -267,8 +267,9 @@ def _read_array(group, dataset_key):
             # cast to a numpy array to avoid multiple calls to
             # _decode_chunk in zarr (or h5py)
             data = da.from_array(data, chunks=data.chunks)
-            shape = da.from_array(ragged_shape, chunks=ragged_shape.chunks)
-            shape = shape.rechunk(data.chunks)
+            shape = da.from_array(
+                ragged_shape, chunks=data.chunks
+            )  # same chunks as data
```
ericpre, line 272: Update the comment to also mention the reason?

ericpre:
Ping @CSSFrancis! 😉 

Just to avoid conflicts with #213.

CSSFrancis:
> Ping @CSSFrancis! 😉
> 
> Just to avoid conflicts with #213.

@ericpre Oops let me add that comment and then we can merge.  This has been working for the last week (at least well enough that I forgot about coming back to the fix :))

ericpre:
And a changelog entry please!

ericpre (approved):

Issue #212, opened 2024-01-18
Error when saving markers with navigation dimension to hspy format


magnunor:
When making a `Texts` marker with navigation dimensions, and adding to a signal with navigation dimensions. An error is raised when trying to save it (plotting it works fine):

## Minimal working example

```python
import numpy as np
import hyperspy.api as hs
from hyperspy.drawing._markers.texts import Texts

s = hs.signals.Signal2D(np.zeros((2, 100, 100)))
text_offset = np.empty(2, dtype=object)
text = np.empty(2, dtype=object)

for i in range(2):
    text_offset[i] = [[50, 40], ]
    text[i] = ['(30.0, -15.0)']

marker_texts = Texts(offsets=text_offset, texts=text, color="red")
s.add_marker(marker_texts, permanent=True)
s.save("testsignal.hspy")
```

Gives the error:

```python
rsciio/hspy/_api.py:121, in HyperspyWriter._get_object_dset(group, data, key, chunks, **kwds)
    119     dtype = data[test_ind].compute().dtype
    120 else:
--> 121     dtype = data[test_ind].dtype
    122 dset = group.require_dataset(
    123     key, data.shape, dtype=h5py.special_dtype(vlen=dtype), chunks=chunks, **kwds
    124 )
    125 return dset

AttributeError: 'list' object has no attribute 'dtype'
```
## Python environment:

- HyperSpy version: 2.0.0
- rsciio version: 0.3

## Additional context

This does not occur if there are no navigation dimensions.

ericpre:
I had a quick look at this and the issue is that h5py doesn't support for these numpy dtype (for example: https://docs.h5py.org/en/stable/strings.html#what-about-numpy-s-u-type). The way around these is to convert the unsupported type to something supported by h5py.

For example, the following example works for saving without changing rosettasciio code:

```python
import hyperspy.api as hs
import numpy as np

# Create a Signal2D with 1 navigation dimension
rng = np.random.default_rng(0)
data = np.ones((10, 100, 100))
s = hs.signals.Signal2D(data)

s2 = hs.signals.Signal2D(data)

import h5py
dt = h5py.special_dtype(vlen=str) 

offsets = np.empty(s.axes_manager.navigation_shape, dtype=object)
texts = np.empty(s.axes_manager.navigation_shape, dtype=object)


for index in np.ndindex(offsets.shape):
    i = index[0]
    offsets[index] = rng.random((10, 2))[:i+2] * 100
[...]

PR #213, opened 2024-01-21
Fix saving ragged array string hspy

ericpre:
Fix for #212.

### Progress of the PR
- [ ] add tests,
- [ ] ready for review.

```python
"""
import numpy as np
import hyperspy.api as hs
import h5py

fname = "test.hspy"

rng = np.random.default_rng(0)
data = np.ones((5, 100, 100))
s = hs.signals.Signal2D(data)

# Create an navigation dependent (ragged) Texts marker
offsets = np.empty(s.axes_manager.navigation_shape, dtype=object)
texts = np.empty(s.axes_manager.navigation_shape, dtype=object)

for index in np.ndindex(offsets.shape):
    i = index[0]
    offsets[index] = rng.random((5, 2))[: i + 2] * 100
    texts[index] = np.array(
        ["a" * (i + 1), "b", "c", "d", "e"][: i + 2]
    )

m = hs.plot.markers.Texts(
    offsets=offsets,
    texts=texts,
    sizes=3,
    facecolor="black",
)

s.add_marker(m, permanent=True)
s.plot()
s.save(fname, overwrite=True)

s2 = hs.load(fname)

m_texts = m.kwargs['texts']
m2_texts = s2.metadata.Markers.Texts.kwargs['texts']


for index in np.ndindex(m_texts.shape):
    np.testing.assert_equal(m_texts[index], m2_texts[index])

```

### DIFF
--- a/rsciio/_hierarchical.py
+++ b/rsciio/_hierarchical.py
@@ -670,6 +670,11 @@ class HierarchicalWriter:
     def _store_data(*arg):  # pragma: no cover
         raise NotImplementedError("This method must be implemented by subclasses.")
 
+    @staticmethod
+    def _convert_ragged_string_array(data):
+        # If not implemented by subclass, return input
+        return data
+
     @classmethod
     def overwrite_dataset(
         cls,
@@ -720,9 +725,7 @@ class HierarchicalWriter:
             # Saving numpy unicode type is not supported in h5py
             data = data.astype(np.dtype("S"))
 
-        if data.dtype == np.dtype("O"):
-            dset = cls._get_object_dset(group, data, key, chunks, **kwds)
-        else:
+        if data.dtype != np.dtype("O"):
             got_data = False
             while not got_data:
                 try:
@@ -760,9 +763,15 @@ class HierarchicalWriter:
                 new_data = np.empty(shape=data.shape, dtype=object)
                 shapes = np.empty(shape=data.shape, dtype=object)
                 for i in np.ndindex(data.shape):
-                    new_data[i] = data[i].ravel()
-                    shapes[i] = np.array(data[i].shape)
+                    data_ = data[i]
+                    if np.issubdtype(data_.dtype, np.dtype("U")):
+                        # h5py doesn't support numpy array with unicode dtype
+                        new_data[i] = cls._convert_ragged_array_string(data_.ravel())
+                    else:
+                        new_data[i] = data_.ravel()
+                    shapes[i] = np.array(data_.shape)
 
+            dset = cls._get_object_dset(group, new_data, key, chunks, **kwds)
             shape_dset = cls._get_object_dset(
                 group, shapes, f"_ragged_shapes_{key}", shapes.shape, **kwds
             )
--- a/rsciio/_hierarchical.py
+++ b/rsciio/_hierarchical.py
@@ -147,8 +147,6 @@ class HierarchicalReader:
         self.version = self.get_format_version()
         self.Dataset = None
         self.Group = None
-        self.unicode_kwds = None
-        self.ragged_kwds = None
 
         if self.version > Version(version):
             warnings.warn(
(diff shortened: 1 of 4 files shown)


rsciio/_hierarchical.py
```diff
@@ -52,9 +52,20 @@ def flatten_data(x):
 
 
 def unflatten_data(data, shape):
+    try:
+        # Since h5py doesn't support numpy unicode dtype, we need to save
+        # data with h5py variable length str dtype and when reading the
+        # ragged data, we need to convert it back to numpy unicode dtype
+        if data.dtype.metadata["vlen"].metadata["vlen"] == str:
```
ericpre, line 59: I am not sure how robust it would be but this is the only way (workaround) that I could find to figure out if the dtype of the ragged are string.

ericpre:
@CSSFrancis, can you please have a look at this PR?
An alternative would be to save an attributes which contain the dtype for each array of the ragged array and convert it back when reading - similarly as it is done with reshaping ragged array. I didn't look much how easy it would be but I thought about it as being more robust/generic.

rsciio/zspy/_api.py
```diff
@@ -109,6 +105,12 @@ def _get_object_dset(group, data, key, chunks, **kwds):
             dtype = data[test_ind].compute().dtype
         else:
             dtype = data[test_ind].dtype
+        if np.issubdtype(dtype, str):
+            size_of_char = np.dtype("U1").itemsize
+            # Go through the whole array to find the largest string type
+            for index in np.ndindex(data.shape):
+                if data[index].dtype.itemsize // size_of_char > dtype.itemsize:
+                    dtype = data[index].dtype
```
CSSFrancis, line 113: How does this work with lazy arrays of strings?  I don't think that this should work? If it does it won't be very efficient as you have to look at every single index. 

We don't __have__ to support saving lazy arrays of strings but we should at least document it...  Part of me is just tempted to pre allocate something like 128 characters but as soon as I say that then something is going to come up where I want more than 128.

CSSFrancis (commented):
Hmm I think that this is a good solution to hard problem.  I don't love that you have to go through every position to find the longest string.  I almost want to just say preallocate 128 characters and call it good.  It might be a waste but storage is cheap.   We could have a setting where it either checks or goes with the default.

That or we could unwrap and wrap the data which might be the best way to do things.

rsciio/zspy/_api.py
ericpre: Yes, I have a look at the feasibility of the alternative approach mentioned in https://github.com/hyperspy/rosettasciio/pull/213#issuecomment-1902630331, because if this is possible, this would be done in the `flatten`/`unflatten` functions and therefore be lazy friendly.

rsciio/_hierarchical.py
ericpre: Yeah, I am not keen on this but I would ok with keeping it knowing that I couldn't find any another solution with h5py, even if there is a risk that it breaks at some point because this attribute is "weakly" documented.
This is another justification for consider the alternative approach more seriously.

rsciio/zspy/_api.py
CSSFrancis: I think to do that you would basically have to store everything as 1 long string:

For example 

```["a", "b", "cd", "e"] --> "abcde"``` and then also save an int array `[1,1,2,1]`

You could also use a deliminator:
```["a", "b", "cd", "e"] --> "a-b-cd-e"```


The flatten function:

```
def flatten_strings(arr):
    string_lengths = np.empty(arr.shape, dtype=object)
    flat_strings= np.empty(arr.shape, dtype=object)
    for i in np.ndindex(arr):
        string_lengths[I] = np.char.str_len(arr)
        flat_strings[i] =  "".join( arr[I])
    return string_lengths, flat_strings
```

CSSFrancis: Note that 2-d ragged arrays of strings might be more of a headache so I'm okay with sticking to the 1d case.

ericpre:
Done in #217.

PR #217, opened 2024-01-26
Fix saving ragged array string hspy - take 2

ericpre:
Fix for #212 with small variations of #213: for zspy, use the msgpack encoder that can more deal with strings.

### Progress of the PR
- [x] Fix saving ragged array of strings in `hspy` and `zspy` format,
- [x] add tests,
- [x] ready for review.

```python
import numpy as np
import hyperspy.api as hs

fname = "test.hspy"

rng = np.random.default_rng(0)
data = np.ones((5, 100, 100))
s = hs.signals.Signal2D(data)

# Create an navigation dependent (ragged) Texts marker
offsets = np.empty(s.axes_manager.navigation_shape, dtype=object)
texts = np.empty(s.axes_manager.navigation_shape, dtype=object)

for index in np.ndindex(offsets.shape):
    i = index[0]
    offsets[index] = rng.random((5, 2))[: i + 2] * 100
    texts[index] = np.array(
        ["a" * (i + 1), "b", "c", "d", "e"][: i + 2]
    )

m = hs.plot.markers.Texts(
    offsets=offsets,
    texts=texts,
    sizes=3,
    facecolor="black",
)

s.add_marker(m, permanent=True)
s.plot()
s.save(fname, overwrite=True)

s2 = hs.load(fname)

m_texts = m.kwargs['texts']
m2_texts = s2.metadata.Markers.Texts.kwargs['texts']


for index in np.ndindex(m_texts.shape):
    np.testing.assert_equal(m_texts[index], m2_texts[index])
```

### DIFF
--- a/rsciio/zspy/_api.py
+++ b/rsciio/zspy/_api.py
@@ -79,16 +79,11 @@ class ZspyReader(HierarchicalReader):
 class ZspyWriter(HierarchicalWriter):
     target_size = 1e8
     _file_type = "zspy"
+    _unicode_kwds = {"dtype": object, "object_codec": numcodecs.JSON()}
 
     def __init__(self, file, signal, expg, **kwargs):
         super().__init__(file, signal, expg, **kwargs)
         self.Dataset = zarr.Array
-        self.unicode_kwds = {"dtype": object, "object_codec": numcodecs.JSON()}
-        self.ragged_kwds = {
-            "dtype": object,
-            "object_codec": numcodecs.VLenArray(signal["data"][0].dtype),
-            "exact": True,
-        }
 
     @staticmethod
     def _get_object_dset(group, data, key, chunks, **kwds):
--- a/rsciio/zspy/_api.py
+++ b/rsciio/zspy/_api.py
@@ -22,6 +22,7 @@ from collections.abc import MutableMapping
 import dask.array as da
 from dask.diagnostics import ProgressBar
 import numcodecs
+import numpy as np
 import zarr
 
 from rsciio._docstrings import (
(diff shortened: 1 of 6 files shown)


ericpre:
@CSSFrancis, can you have a look at this PR please?

rsciio/zspy/_api.py
```diff
@@ -102,17 +98,32 @@ def _get_object_dset(group, data, key, chunks, **kwds):
...
         these_kwds = kwds.copy()
         these_kwds.update(dict(dtype=object, exact=True, chunks=chunks))
-        test_ind = data.ndim * (0,)
-        # Need to know the underlying dtype for the codec
-        # Note this can't be an object array
-        if isinstance(data, da.Array):
-            dtype = data[test_ind].compute().dtype
+
+        if dtype is None:
+            test_data = data[data.ndim * (0,)]
```
CSSFrancis, line 103: Can we just compute the data and then check the sub array type with the metadata like above? It seems slightly better than what we are doing now?

CSSFrancis (approved):
@ericpre This looks much better!  It seems like we just needed to read the zarr documentation a little more :)

rsciio/zspy/_api.py
ericpre: Sorry, I am not sure if I follow what you are referring to.

CSSFrancis: Anyways this looks good!

ericpre: > I just always feel like there should be a better way of doing this from a numpy object array but that doesn't seem like it is the case. We really want to be able to set a constant dtype for the ragged array. The current implementation is just too flexible for my liking.
> 
> That's a bit of an aside, I think if we really wanted to solve this problem we would have to add something like [awkwardarray](https://awkward-array.org/doc/main/) as a dependency.

Yes quite likely. For now, this fix should do, hopefully! ;)

PR #222, opened 2024-02-02
Add osx arm64 GitHub ci

ericpre:
Similar as https://github.com/hyperspy/hyperspy/pull/3305.

https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/
https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/

### Progress of the PR
- [x] Add osx arm64 build to GitHub CI,
- [x] Speed running test suite using all available CPUs (3 or 4) instead of only 2,
- [x] Remove unnecessary  parametrisation of hspy/zspy ragged signal test which was slowing down the test suite,
- [n/a] add tests,
- [x] ready for review.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -69,7 +69,19 @@ jobs:
         name: Install Python
         with:
           python-version: ${{ matrix.PYTHON_VERSION }}
-
+        
+      - name: Get the number of CPUs
+        id: cpus
+        run: |
+          import os, platform
+          num_cpus = os.cpu_count()
+          print(f"Number of CPU: {num_cpus}")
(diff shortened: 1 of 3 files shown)


ericpre:
I couldn't find any information on when the `macos-14` (using arm chips) will replace the `macos-latest` (using intel chips). I would expect that the intel runner will be around for a while, which also mean that once in a while we may need to update  the `macos-14` to a new version.
We will need to test both architectures for a while.

jlaehne:
OK, I just think we should also be somewhat *economic* on resources of test runners and try not to escalate the instances too much.

jlaehne (approved):

Issue #219, opened 2024-01-28
Support more features of the mib reader


ericpre:
Support for mib files (quantum detector) has been added in #174 and this is a list of features that would be adding:

- [ ] infer navigation shape from flyback - reading exposure time is implemented but test files are needed.
- [ ] add support for reading ROI
- [ ] add support for reading RAW data

magnunor:
I'm very interested all of these, if anyone needs test files we can definitely supply that. Maybe also a pull request if someone has time.

ericpre:
As it says the first one should be very easy, the two others would need a volunteer!

M0hsend:
Many thanks for adding this! @ericpre 
Just to note that you seem to have two methods to read mib files now, one `file_reader` and one `load_mib_data`. The `file_reader` is applying a up-down flip on the diffraction patterns [here](https://github.com/hyperspy/rosettasciio/blob/5a441b8a47d7899560c078a61c23fd324c093554/rsciio/quantumdetector/_api.py#L540C12-L541C1). Would be good to add this to the metadata.
Also just wondering about the plans for adding the cross gap for quad configurations. Will this be re-added to pyxem?

CSSFrancis:
> Many thanks for adding this! @ericpre Just to note that you seem to have two methods to read mib files now, one `file_reader` and one `load_mib_data`. The `file_reader` is applying a up-down flip on the diffraction patterns [here](https://github.com/hyperspy/rosettasciio/blob/5a441b8a47d7899560c078a61c23fd324c093554/rsciio/quantumdetector/_api.py#L540C12-L541C1). Would be good to add this to the metadata. Also just wondering about the plans for adding the cross gap for quad configurations. Will this be re-added to pyxem?

@M0hsend We can definitely add this back to pyxem!  I just need to understand what a good workflow would be and what information is/isn't available.  I don't have/use a Merlin detector so supporting some of these features is a bit difficult without a bit more direction :). 

Is the cross gap information stored in the metadata? Is the cross gap information constant or could you arrange the detectors anyway you want?  I think we would want something like an `map_to_detector` method for pyxem which would add pixels with np.nan values at a certain fraction of the dataset where the cross is.  We could think about creating a `Dectector` class which defines a layout for a certain detector and then maps certain pixels from the original to the transformed dataset.  That gives us much more freedom to start thinking about how to handle any possible detector shape.

M0hsend:
[...]

PR #235, opened 2024-03-16
Add support for pixel trigger acquisition in quantumdetector reader

ericpre:
Fix first item of https://github.com/hyperspy/rosettasciio/issues/219.

### Progress of the PR
- [x] Use timestamps to get navigation shape when the navigation shape is not available otherwise (acquisition with pixel trigger and scanshape not in metadata),
- [x] update docstring,
- [x] update user guide,
- [x] add tests,
- [x] ready for review.

### DIFF
--- a/doc/user_guide/supported_formats/quantumdetector.rst
+++ b/doc/user_guide/supported_formats/quantumdetector.rst
@@ -9,17 +9,6 @@ store a series of diffraction patterns from scanning transmission electron
 diffraction measurements. It supports reading data from camera with one or
 four quadrants.
 
-If a ``hdr`` file with the same file name was saved along the ``mib`` file,
-it will be used to infer the navigation shape of the providing that the option
-"line trigger" was used for the acquisition. Alternatively, the navigation
-shape can be specified as an argument:
-
-.. code-block:: python
-
-    >>> from rsciio.quantumdetector import file_reader
-    >>> s_dict = file_reader("file.mib", navigation_shape=(256, 256))
-
-
 API functions
 ^^^^^^^^^^^^^
 
--- a/rsciio/quantumdetector/_api.py
+++ b/rsciio/quantumdetector/_api.py
@@ -327,13 +327,22 @@ def load_mib_data(
     data = data["data"]
     if not return_mmap:
         if lazy:
-            data = da.from_array(data, chunks=chunks)
+            if isinstance(chunks, tuple) and len(chunks) > 2:
+                # Since the data is reshaped later on, we set only the
+                # signal dimension chunks here
+                _chunks = ("auto",) + chunks[-2:]
+            else:
+                _chunks = chunks
+            data = da.from_array(data, chunks=_chunks)
         else:
             data = np.array(data)
 
     # remove navigation_dimension with value 1 before reshaping
     navigation_shape = tuple(i for i in navigation_shape if i > 1)
     data = data.reshape(navigation_shape + mib_prop.merlin_size)
+    if lazy and isinstance(chunks, tuple) and len(chunks) > 2:
+        # rechunk navigation space when chunking is specified as a tuple
+        data = data.rechunk(chunks)
 
     if return_headers:
         return data, headers
--- a/rsciio/quantumdetector/_api.py
+++ b/rsciio/quantumdetector/_api.py
@@ -22,6 +22,7 @@
 import logging
 import os
 from pathlib import Path
+import warnings
 
 import dask.array as da
 import numpy as np
@@ -130,14 +131,14 @@ class MIBProperties:
                     self.file_size = f.tell()
                     self.buffer = False
                     self.path = path
-            except:  # pragma: no cover
+            except BaseException:  # pragma: no cover
                 raise RuntimeError("File does not contain MIB header.")
         elif isinstance(path, bytes):
             try:
                 head = path[:384].decode().split(",")
                 self.file_size = len(path)
                 self.buffer = True
-            except:  # pragma: no cover
+            except BaseException:  # pragma: no cover
                 raise RuntimeError("Buffer does not contain MIB header.")
         else:  # pragma: no cover
             raise TypeError("`path` must be a str or a buffer.")
@@ -401,7 +402,7 @@ def parse_exposures(headers, max_index=10000):
     from the headers. By default, reads only the first 10 000 frames.
 
     >>> from rsciio.quantumdetector import load_mib_data, parse_exposures
-    >>> data, headers = load_mib_data(path, return_header=True, return_mmap=True)
+    >>> data, headers = load_mib_data(path, return_headers=True, return_mmap=True)
     >>> exposures = parse_exposures(headers)
 
     All frames can be parsed by using ``max_index=-1``:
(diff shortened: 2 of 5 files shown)


ericpre:
@magnunor, since you mentioned that you are "very interested" in this feature (https://github.com/hyperspy/rosettasciio/issues/219#issuecomment-1913672352), can you please review it? If not, I will merge it.

magnunor:
@ericpre, very nice! I'll have a look at this, and test it on some of my datasets. Probably tomorrow.

magnunor:
I tested this on a number of datasets, and it seems to work nicely! In the current `main` branch, it gives the shape `(65536, 256, 256)`, while with this pull request it gives `(256, 256, 256, 256)`.

magnunor:
I have some more feedback, but I won't have time to do it before Thursday.

magnunor (commented):
One thing which is missing is either i) a better error message when the navigation shape is missing, and the automatic navigation-shape-guesser doesn't work, or ii) better handling of mib-files where the navigation shape is unknown.

For example, I have file where the HDR-file says `'Frames in Acquisition (Number)': '0'`, and where there are no flyback pixels. Currently the error message is a bit cryptic: `ValueError: cannot reshape array of size 0 into shape (256,256)`

For ii), it should be possible to estimate the number of frames based on the size of the file, header size, number of chips and dtype. It could then return a 3D array.

doc/user_guide/supported_formats/quantumdetector.rst
```diff
@@ -9,17 +9,6 @@ store a series of diffraction patterns from scanning transmission electron
...
 
-If a ``hdr`` file with the same file name was saved along the ``mib`` file,
-it will be used to infer the navigation shape of the providing that the option
-"line trigger" was used for the acquisition. Alternatively, the navigation
-shape can be specified as an argument:
-
-.. code-block:: python
-
-    >>> from rsciio.quantumdetector import file_reader
-    >>> s_dict = file_reader("file.mib", navigation_shape=(256, 256))
```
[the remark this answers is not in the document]
ericpre: I moved it to the docstring and it is still in the same page of the user guide: https://rosettasciio--235.org.readthedocs.build/en/235/user_guide/supported_formats/quantumdetector.html#rsciio.quantumdetector.file_reader

ericpre:
Thanks @magnunor for the review. If this PR doesn't introduce any regression, I think that it would be useful to merge it to include it in the next release.
I replied to your comments below, some of these may need more time to understand and document the use case and get small test files. I will try to deal with the `'Frames in Acquisition (Number)': '0'` case, as this should be fairly straightforward even if this is not clear why it would occur.

> One thing which is missing is either i) a better error message when the navigation shape is missing, and the automatic navigation-shape-guesser doesn't work, or ii) better handling of mib-files where the navigation shape is unknown.

Most of the time, the `navigation_shape` is not known. It should be included explicitly in the `hdr` file generated by recent version of the software (assuming that this feature has been released).
Are you aware of there other cases where estimating the navigation shape doesn't work (other than the one that you already mentioned: the `'Frames in Acquisition (Number)': '0'` case)?

> For example, I have file where the HDR-file says `'Frames in Acquisition (Number)': '0'`, and where there are no flyback pixels. Currently the error message is a bit cryptic: `ValueError: cannot reshape array of size 0 into shape (256,256)`

Do you know in what situation does this occur? I am wondering if there is some buggy/corrupted file or if this actually corresponding to a specific use case. If so, it would be useful to add a comment about it in the code.

Assuming this should be with single frame files, there are already some of these file in the test suite and the tests are still passing?

> For ii), it should be possible to estimate the number of frames based on the size of the file, header size, number of chips and dtype. It could then return a 3D array.

The number of frames is already known from the code, how do you think this information should be used for reshaping purposes?

ericpre:
@magnunor, the `'Frames in Acquisition (Number)': '0'` case should now be covered, do you want to have another look?

magnunor:
> Thanks @magnunor for the review. If this PR doesn't introduce any regression, I think that it would be useful to merge it to include it in the next release.

Based on testing this PR on a number of datasets, I don't think there are any regressions.

> Are you aware of there other cases where estimating the navigation shape doesn't work (other than the one that you already mentioned: the 'Frames in Acquisition (Number)': '0' case)?

One case could be where flyback was not used, meaning all the pixels have the same frame time.

Another case could be where the acquisition was interrupted. On some older datasets, it seems like the HDR file has a higher number ('Frames in Acquisition (Number)':) than the actual data in the MIB-file. Here is an example of such a file: https://filesender.sikt.no/?s=download&token=91e8ff4b-dda8-4cb8-997b-0504d7edaca2 (the link will expire 20. april)


> Do you know in what situation does this occur? I am wondering if there is some buggy/corrupted file or if this actually corresponding to a specific use case. If so, it would be useful to add a comment about it in the code.
> 
> Assuming this should be with single frame files, there are already some of these file in the test suite and the tests are still passing?

I am actually not sure when the 'Frames in Acquisition (Number)': '0' occurs. The file is not single-frame. This is a dataset from 2018, so an older version of the Merlin software.

> The number of frames is already known from the code, how do you think this information should be used for reshaping purposes?

Looking a bit closer at this, it seems like the part which is checking for the number of frames in the `load_mib_data` is not catching some of the cases where the acquisition is interrupted. Essentially, this part of the code is not being used for the file linked earlier: https://github.com/hyperspy/rosettasciio/blob/4f2d78ec41d5e7eedef7874bff21cd17a0ae13a4/rsciio/quantumdetector/_api.py#L276

rsciio/quantumdetector/_api.py
```diff
@@ -517,13 +544,43 @@ def file_reader(
...
+
+            times_diff = np.diff(times).astype(float)
+            if len(times_diff) > 0:
+                # Substract the mean and take the first position above 0
+                indices = np.argwhere(times_diff - np.mean(times_diff) > 0)
+                if len(indices) > 0 and len(indices[0]) > 0:
+                    frame_per_trigger = indices[0][0] + 1
+
+        if frames_number == 0:
+            # Some hdf file file have the "Frames per Trigger (Number)": 0
```
magnunor, line 578:
```suggestion
            # Some hdr files have the "Frames per Trigger (Number)": 0
```

ericpre:
> > Are you aware of there other cases where estimating the navigation shape doesn't work (other than the one that you already mentioned: the 'Frames in Acquisition (Number)': '0' case)?
> 
> One case could be where flyback was not used, meaning all the pixels have the same frame time.

Okay, but in that case, this is no way to estimate the navigation shape and it needs to be provided by the users. By default the current implementation will return a navigation_shape of (number_of_frames, ).

> Another case could be where the acquisition was interrupted. On some older datasets, it seems like the HDR file has a higher number ('Frames in Acquisition (Number)':) than the actual data in the MIB-file. Here is an example of such a file: https://filesender.sikt.no/?s=download&token=91e8ff4b-dda8-4cb8-997b-0504d7edaca2 (the link will expire 20. april)
> 
> > Do you know in what situation does this occur? I am wondering if there is some buggy/corrupted file or if this actually corresponding to a specific use case. If so, it would be useful to add a comment about it in the code.
> > Assuming this should be with single frame files, there are already some of these file in the test suite and the tests are still passing?
> 
> I am actually not sure when the 'Frames in Acquisition (Number)': '0' occurs. The file is not single-frame. This is a dataset from 2018, so an older version of the Merlin software.

I checked the merlin manual, it seems that it may correspond to a case where the acquisition is set to be "continuous and indefinite". The wording is not exactly as the one of used in the hdr file but it seems to make sense to a certain extent!

> > The number of frames is already known from the code, how do you think this information should be used for reshaping purposes?
> 
> Looking a bit closer at this, it seems like the part which is checking for the number of frames in the `load_mib_data` is not catching some of the cases where the acquisition is interrupted. Essentially, this part of the code is not being used for the file linked earlier:
> 
> https://github.com/hyperspy/rosettasciio/blob/4f2d78ec41d5e7eedef7874bff21cd17a0ae13a4/rsciio/quantumdetector/_api.py#L276
> 
> However, these issues also happen in the current `main` branch. So this is not a regression.

This was a bug in the code to reshape in case of interrupted acquisition - this should be fixed now. Can you confirm that the dataset you shared here have 5 pixels per line?

rsciio/tests/test_quantumdetector.py
```diff
@@ -363,3 +374,23 @@ def test_load_save_cycle(tmp_path):
     assert s.axes_manager.navigation_shape == s2.axes_manager.navigation_shape
     assert s.axes_manager.signal_shape == s2.axes_manager.signal_shape
     assert s.data.dtype == s2.data.dtype
+
+
+def test_Frames_in_Acquisition_zero():
```
magnunor, line 379: I don't think upper-case should be used in function names

magnunor:
This should be ready for merging now! Very nice improvements to the MIB-reader :)

----------------

> Okay, but in that case, this is no way to estimate the navigation shape and it needs to be provided by the users. By default the current implementation will return a navigation_shape of (number_of_frames, ).

Indeed! Ignore the things I was writing about that, due to the bug you fixed this is no longer relevant.

> This was a bug in the code to reshape in case of interrupted acquisition - this should be fixed now. Can you confirm that the dataset you shared here have 5 pixels per line?

I think for this specific dataset, things are a bit strange. These frames should be from a single line. But the `times_diff - np.mean(times_diff)` gives:

```python
array([-8.66666667, -8.66666667, -8.66666667, -8.66666667,  1.33333333,
        1.33333333,  1.33333333,  1.33333333,  1.33333333,  1.33333333,
        1.33333333,  1.33333333,  1.33333333,  1.33333333,  1.33333333,
        1.33333333,  1.33333333,  1.33333333,  1.33333333,  1.33333333,
        1.33333333,  1.33333333,  1.33333333,  1.33333333,  1.33333333,
        1.33333333,  1.33333333,  1.33333333,  1.33333333,  1.33333333])
```
But I don't expect the file reader to be able to handle the navigation shape in strange datasets such as these. What was annoying earlier was that the data wasn't loading at all, but with the bug fix it loads just nicely :)

ericpre:
Okay, thank you, I fixed the typos in the test name, we should be all good now!

Issue #243, opened 2024-03-25
Improve units conversion in EMD velox files


ThorstenBASF:
#### Describe the bug
If an image has a different number of pixels in X and Y, it can happen that axes are set to different units.
In my case I had an image of 256x512 with an original pixelsize of 5.1e-09m.
This is converted to 5.1nm in X and 0.0051µm in Y.

This leads to an error while I want to save the image with a scalebar, due the sanity check of the file_writer.
(see io_plugins/image.py, line 147)

#### To Reproduce
(The testfile contains confidential data, so i can't share, if nessessary I could creat one with public data.)
```
import hyperspy.api as hs
myfile = "testfile.emd"
s = hs.load(myfile, select_type='images')
s[0].save(myfile[:-4])+".jpg"), scalebar=True)
```

#### Expected behavior
Not sure where to fix it. I think best would be to have same units even if X Y has a ratio of 1:2, but here I didn't checked(found) the code.
The santiy check could also be extended to compare different units, but that would be much more effort.

#### Python environment:
 - HyperSpy version: 1.7.5 AND 2.0.1
 - Python version: 3.11.6

#### Additional context
Images of original_metadata and axes.
As a workaround I just set [1]=[0] for scale and units.

![original_metadata](https://github.com/hyperspy/hyperspy/assets/109654649/7a6951b4-fcaa-49f2-acec-af94b6c6909d)
![2d_signal_hyperspy](https://github.com/hyperspy/hyperspy/assets/109654649/a9b257ea-0a1c-41b9-962e-6f388f6cee1f)

ericpre:
@ThorstenBASF for opening this issue. Indeed, even if the current behaviour is still correct, it is not convenient and it would be better to keep the same units when possible (same dimension).
The issue is that when loading emd velox file, the units are converted automatically for conveninence (for example: to get units in "nm" or "um" instead of "m", which is the units saved in the emd file for the calibration) but axes are converted independently and it is possible that in some corner cases, there end up being different.

This could be improved by checking that both units are different and pick one of the two. What criterion should we use to pick one of the two?
@ThorstenBASF, in your specific case, which one of the two ("nm" or "um") would make more sense?
I can't think of an objective criterion and I think this is arbitrary. The main impact is to have small or large number in the display of the scalebar. Also it we consider large aspect ratio which is not uncommon, then I suspect that it wouldn't be possible to find a criterion that fit all corner cases!

[...]

PR #248, opened 2024-04-05
fix for issue #243 (EMD axes data)

ThorstenBASF:
### Description of the change
X and Y could have different units calculated by "_convert_scale_units" now the units itself is only set once and other axe values are set this unit. Different units results in an error of an sanity check of matplotlib sclaebar. And the offsets could be wrong. Changes:
- the units for scale_x are fixed to be used for scale_y and offsets to always have same units for all.
- made a own def to do this:  "_convert_scale_units_to"

### Progress of the PR
- [ ] add tests,
- [ ] ready for review.

```
import hyperspy.api as hs
import numpy as np

myfile = "_test_SI.emd"
s = hs.load(myfile, select_type='images')

s[0].data -= np.min(s[0].data)
s[0].data /= np.max(s[0].data)
s[0].data *= 255
s[0].data = s[0].data.astype(np.uint8)
s[0].save(((myfile[:-4])+".jpg"), scalebar=True, overwrite=True)
```

### DIFF
--- a/rsciio/emd/_emd_velox.py
+++ b/rsciio/emd/_emd_velox.py
@@ -358,14 +358,14 @@ class FeiEMDReader(object):
         scale_x = self._convert_scale_units(
             pix_scale["width"], original_units, data.shape[i + 1]
         )
-        scale_y = self._convert_scale_units(
-            pix_scale["height"], original_units, data.shape[i]
+        scale_y = self._convert_scale_units_to(
+            pix_scale["height"], original_units, scale_x[1]
         )
-        offset_x = self._convert_scale_units(
-            offsets["x"], original_units, data.shape[i + 1]
+        offset_x = self._convert_scale_units_to(
+            offsets["x"], original_units, scale_x[1]
         )
-        offset_y = self._convert_scale_units(
-            offsets["y"], original_units, data.shape[i]
+        offset_y = self._convert_scale_units_to(
+            offsets["y"], original_units, scale_x[1]
         )
         axes.extend(
             [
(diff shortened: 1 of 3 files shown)


ThorstenBASF:
Not sure, if units can ever be _None_, but _convert_scale_units also checks for _None_.
Problem is, this would result in an error.

ericpre:
Another thing: it would be better to have a small test file (most likely need to make one for this specific purpose) but it this is not straightforward and may need some time, this is not a big deal and we can merge this PR without it.

ericpre:
Yes, this is good. Can you add a changelog entry as explained in https://github.com/hyperspy/rosettasciio/blob/main/upcoming_changes/README.rst?

ThorstenBASF:
A file for the changelog has been added. I checked other pull requests, there was always a "+" in front of the text. I didn't find anything about it, and I tried "towncrier" and the "+" felt wrong. So, I removed it, let me know if it is necessary.

```
0.3 (2024-04-08)
================

Bug Fixes
---------

- + Fix to avoid mismatching units for scale and offset reading ``emd`` files. (`#243 <https://github.com/hyperspy/rosettasciio/issues/243>`_)
```

Thank you very much for your help and guidance.

ThorstenBASF:
Just found another incompatibility with matplotlib for SAD images (reciprocal scaling).
Reciprocal unit strings are using whitespaces with the current format.
For example "1/m" vs "1 / m". matplotlib wants the first one, but 2nd one is used.

This could be fixed by use the "short pretty" format instead of the "short" only.
A single letter "P" would fix it.

`        converted_units = "{:~}".format(converted_v.units)
` to
`        converted_units = "{:~P}".format(converted_v.units)
`

Should I add this to current pull request, or open another issue?

ericpre:
@ThorstenBASF: congratulation on your first pull request to rosettasciio! 🎉 

I tweaked the release entry and simplify the code to make it more readable by avoid using list when not necessary.

thorwiec:
> Yes, please, let's do this in a separate PR, I am not sure if I understand exactly what is the issue and generally it is easier to follow discussions when they are short and self-contain.

This issue seems fixed with 2.x (see below). I'm still working with 1.7x (stability reasons). But we'll change to 2.x soon.
https://github.com/hyperspy/rosettasciio/blob/e3e76800baf2981b9c0b1248e94d7dfb7d14eca1/rsciio/image/_api.py#L191-L195

Thanks again, next time I'll will need less advice. I promise! :)

PR #250, opened 2024-04-06
Use ruff

ericpre:
### Progress of the PR
- [x] Use `ruff` and remove `black`,
- [n/a] add tests,
- [x] ready for review.


rsciio/phenom/_api.py
```diff
@@ -788,16 +789,16 @@ def _read_LineScanAnalysis(self, label, am):
...
-        top = self._read_float64()
-        right = self._read_float64()
-        bottom = self._read_float64()
-        color_intensities = self._read_float64s()
+        _ = self._read_float64()  # left
+        _ = self._read_float64()  # top
+        _ = self._read_float64()  # right
+        _ = self._read_float64()  # bottom
+        _ = self._read_float64s()  # color_intensities
         width = self._read_uint32()
```
jlaehne, line 797: Why are the `int` lines handled differently than the `float` lines?

ericpre: This changes is to avoid warning about variable not used, the `width` variable must be used somewhere, while the ones that I changed are not. I have used a placeholder for code readability reason: these calls return something but we don't use it.

jlaehne: Could you add a comment to explain this for future developers?

jlaehne (approved):

Issue #266, opened 2024-05-28
Memory spike when loading .mib data lazily


emichr:
#### Describe the bug
with When loading large .mib data with `hs.load("data.mib", lazy=True)`, RSS memory spikes more than size of dataset. For instance, when loading a 31.3 GB dataset without specifying any chunking, the RSS memory spikes at about 95 GB. With different chunking, the memory spike changes, but is still a problem (e.g. RSS spikes at 73 GB with a `(64, 64, 64, 64)` chunking of a `(800, 320|256,256)` uint16 dataset). This figure shows the RSS memory usage as a function of time when running `hs.load(..., lazy=True, chunks=(64, 64, 64,64))`:
![Figure 7](https://github.com/hyperspy/rosettasciio/assets/16645501/6507a1a6-149e-426f-8c57-f38872d76581)

#### To Reproduce
```
signal = hs.load("data.mib", lazy=True)
```

#### Expected behavior
The RSS memory requirement shouldn't exceed the dataset size on disk, and should be much lower when loaded lazily.

#### Python environement:
 - RosettaSciIO version: 0.4
 - Python version: 3.11.9
 - HyperSpy version: 2.1.0
 - Pyxem version: 0.18.0

#### Additional context
The problem was encountered on a HPC cluster, but I assume the problem will persist on other machines.

CSSFrancis:
@emichr can you try the same thing with chunks of (16,16,256,256)?

emichr:
@CSSFrancis The peak got a little lower, but it is still significant (the time is also significant):
![Figure 2_2](https://github.com/hyperspy/rosettasciio/assets/16645501/4a3169b6-98ce-489d-b50f-75bf8b775cdf)

I just tested on a laptop and there are no issues. So it might be an issue with the cluster (or the environment of course). The environment .yml file from the HPC looks like this: 
[pyxem_hpc.txt](https://github.com/hyperspy/rosettasciio/files/15469225/pyxem_hpc.txt)

It might be an issue better discussed with the particular HPC infrastructure, but I guess other users might get into this issue down the line.

ericpre:
Could it be that `memmap` isn't playing well with the distributed scheduler and the mib reader would need the same as https://github.com/hyperspy/rosettasciio/pull/162?

@emichr, just to check this is using the distributed scheduler?

CSSFrancis:
[...]

PR #267, opened 2024-05-29
Add support for dask distributed scheduler in quantum detector reader

ericpre:
### Progress of the PR
- [x] Add support for dask distributed scheduler in quantum detector reader - similar to #162 and #11,
- [x] add tests,
- [x] ready for review.


ericpre:
Thank you @CSSFrancis, as you suggested in https://github.com/hyperspy/rosettasciio/issues/266#issuecomment-2137908588, #11 was useful to figure out the details with the structured dtype - pretty much copy & paste the relevant bits! 😄

rsciio/quantumdetector/_api.py
```diff
@@ -344,6 +355,10 @@ def load_mib_data(
         data = data.rechunk(chunks)
 
     if return_headers:
+        if distributed:
```
CSSFrancis, line 358: You can still return the header by just setting the `key="header"` for a second `memmap_distributed` call.  It will add some time onto the saving of the dataset as the entire dataset might get loaded into ram with most of it thrown away. 

Really what we should do is add things to a `to_store` context manager and then call:
https://github.com/hyperspy/rosettasciio/blob/31bd677cc4c02c5787ae9b61250bc1431f352cba/rsciio/hspy/_api.py#L111

Only once.  That will merge taskgraphs as necessary and might reduce the time for saving certain signals.  I've thought about it for things like saving lazy markers of possibly creating a `hs.save()` function for handling mulitple signals if you wanted to save multiple parts of some anaylsis efficently.  This is a fairly abstract/higher level concept so maybe it would be seledomly used.

CSSFrancis (approved):
This is actually a much smaller change than I thought it would be :). It looks good to me other than it might be nice to return the header as well although a warning that this could increase the loading time would be useful.

rsciio/quantumdetector/_api.py
ericpre: Yes, this will most likely needed to be done at some point! I opened #269 to track it / add more usecases.

PR #272, opened 2024-06-11
Add support for reading `mib` acquired with a given number of frame to skip per line

### LINKED ISSUE #270: Shape of 4DSTEM .mib data triggered with lineskip=1
#### Describe the bug
When loading 4DSTEM .mib data (hdr file:  
[SPEDSTEM_256x256_8cm_100pct.txt](https://github.com/user-attachments/files/15758275/SPEDSTEM_256x256_8cm_100pct.txt)) with lineskip=1, hyperspy.api.load() throws an IO error: 
[error_msg.txt](https://github.com/user-attachments/files/15758261/error_msg.txt). The error disappears when `navigation_shape=(256,256)` is given, but the resulting signal is distorted:
Figure 3
To load the data correctly with the current version of rsciio, an extra frame must be given in the x-axis of the navigation shape: `navigation_shape=(257, 256)` and this gives
Figure 5
As a workaround this is fine, but maybe some improvements can be made to the reader.

The data is available at: https://drive.google.com/drive/folders/1jFQIKGnT1Ru0faFC_rLzbhrjrVIqx_8q?usp=drive_link.

#### To Reproduce
```
signal = hs.load("SPEDSTEM_256x256_8cm_100pct.mib", lazy=True) #Throws IO error
signal = hs.load("SPEDSTEM_256x256_8cm_100pct.mib", lazy=True, navigation_shape=(256,256)) #Works but resulting signal is distorted.
signal = hs.load("SPEDSTEM_256x256_8cm_100pct.mib", lazy=True, navigation_shape=(257,256)) #Works and resulting signal is correct.
```

#### Expected behavior
The raw .mib data has a total of 65792 frames (as seen in the .hdr file), where $256\times256=65536$ frames contain the actual data (can be seen from the .hdr ScanX and ScanY parameters). Then `hs.load("SPEDSTEM_256x256_8cm_100pct.mib", lazy=True, navigation_shape=(256,256))` should skip every 256th frame (counting from 0) as this is the most usual way of interpreting the navigation shape. It should be possible to get the required information from the .hdr file and thus avoid the need for specifying the navigation shape when reading the data.

#### Python environement:
 - RosettaSciIO version: 0.4
 - Python version: 3.11.9
 - HyperSpy version: 2.1.0
 - Pyxem version: 0.19.0

emichr:
### Description of the change
A few sentences and/or a bulleted list to describe and motivate the change:
- Added a function to read the navigation shape and check for lineskips of a .mib dataset from the corresponding .hdr file. This should help with #270.
- Fixed some typos in various docstrings.

### Progress of the PR
- [ ] add tests,
- [x] ready for review.

```python
from rsciio.quantumdetector._api import parse_hdr_file, _get_navigation_shape_from_hdr
hdr = parse_hdr_file("SPEDSTEM_256x256_8cm_100pct.hdr")
# Your new feature...
navigation_shape = _get_navigation_shape_from_hdr(hdr)
print(navigation_shape)
```
[SPEDSTEM_256x256_8cm_100pct.txt](https://github.com/user-attachments/files/15782627/SPEDSTEM_256x256_8cm_100pct.txt)

### DIFF
--- a/rsciio/quantumdetector/_api.py
+++ b/rsciio/quantumdetector/_api.py
@@ -40,7 +40,6 @@ from rsciio.utils.distributed import memmap_distributed
 
 _logger = logging.getLogger(__name__)
 
-
 _PATH_DOCSTRING = """path : str or bytes
             The path to the ``mib`` file, otherwise the memory buffer
             of the ``mib`` file. Lazy loading is not supported with memory
@@ -281,7 +280,7 @@ def load_mib_data(
             # to load is too small, we can't reshape and we fall back to stack of images
             _logger.warning(
                 "The `navigation_shape` doesn't match the number of frames to load. "
-                "The `navigation_shape` is set to the number of read to read: "
+                "The `navigation_shape` is set to the number of frames to read: "
                 f"({number_of_frames_to_load},)."
             )
             navigation_shape = (number_of_frames_to_load,)
@@ -349,6 +348,10 @@ def load_mib_data(
 
     # remove navigation_dimension with value 1 before reshaping
     navigation_shape = tuple(i for i in navigation_shape if i > 1)
+    if np.prod(navigation_shape) != np.prod(data.shape[:-2]):
+        raise ValueError(
+            f"Cannot load MIB file. Navigation shape {navigation_shape} does not match data size {data.shape}"
+        )
     data = data.reshape(navigation_shape + mib_prop.merlin_size)
     if lazy and isinstance(chunks, tuple) and len(chunks) > 2:
         # rechunk navigation space when chunking is specified as a tuple
@@ -357,7 +360,7 @@ def load_mib_data(
     if return_headers:
         if distributed:
             raise ValueError(
-                "Retuning headers is not supported with `distributed=True`."
+                "Returning headers is not supported with `distributed=True`."
             )
         return data, headers
     else:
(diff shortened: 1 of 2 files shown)


ericpre:
Thanks @emichr. As mentioned in https://github.com/hyperspy/rosettasciio/issues/270#issuecomment-2158637124, the logic needs to take into account interrupted acquisition.

Can you make and add [small test files](https://hyperspy.org/rosettasciio/contributing.html#making-test-data-files) to cover this use case (and different number of frame to skip per line)? 
The test suite fails with this PR:
```python
FAILED rsciio/tests/test_quantumdetector.py::test_interrupted_acquisition - assert (9,) == (4, 2)
FAILED rsciio/tests/test_quantumdetector.py::test_non_square[navigation_shape1] - assert (4, 2) == (8,)
FAILED rsciio/tests/test_quantumdetector.py::test_first_last_frame_8[kwargs0-navigation_shape1] - assert (8,) == (4, 2)
FAILED rsciio/tests/test_quantumdetector.py::test_first_last_frame_8[kwargs1-navigation_shape1] - assert (8,) == (4, 2)
FAILED rsciio/tests/test_quantumdetector.py::test_first_last_frame_8[kwargs2-navigation_shape1] - assert (8,) == (4, 2)
FAILED rsciio/tests/test_quantumdetector.py::test_first_last_frame_8[kwargs3-navigation_shape1] - assert (8,) == (4, 2)
FAILED rsciio/tests/test_quantumdetector.py::test_first_last_frame_8[kwargs4-navigation_shape1] - assert (8,) == (4, 2)
FAILED rsciio/tests/test_quantumdetector.py::test_first_last_frame_8[kwargs5-navigation_shape1] - assert (8,) == (4, 2)
FAILED rsciio/tests/test_quantumdetector.py::test_first_last_frame_8[kwargs6-navigation_shape1] - assert (8,) == (4, 2)
FAILED rsciio/tests/test_quantumdetector.py::test_navigation_shape_list_error - Failed: DID NOT RAISE <class 'TypeError'>
```

emichr:
> Thanks @emichr. As mentioned in [#270 (comment)](https://github.com/hyperspy/rosettasciio/issues/270#issuecomment-2158637124), the logic needs to take into account interrupted acquisition.
> 
> Can you make and add [small test files](https://hyperspy.org/rosettasciio/contributing.html#making-test-data-files) to cover this use case (and different number of frame to skip per line)? The test suite fails with this PR:

Sure, I'll look into it as soon as possible!

emichr:
@ericpre There was a bug in my PR that caused the tests to fail. I fixed the bug and it should run fine now - sorry for not checking sooner!

I plan on making some test data soon with the updated Merlin software with various linetriggers, but it will take some time.

ericpre:
This sounds good. Are you okay with including the test file as part of the PR? I would like to review with the test file. When you make the test file, can you also make one with interrupted acquisition?

Would it make sense to actually skip the frame(s) (by slicing the data array in the right place?), instead of increasing the navigation shape?

emichr:
> This sounds good. Are you okay with including the test file as part of the PR? I would like to review with the test file. When you make the test file, can you also make one with interrupted acquisition?

I probably won't be able to make the test data files this week, so the tests will have to wait a bit. I'm okay with waiting for those and including them in the PR when ready.

> Would it make sense to actually skip the frame(s) (by slicing the data array in the right place?), instead of increasing the navigation shape?

Yes, that's a good idea. It required very little changes, and I think I found the right place to do it. Let me know if it's not the best way of doing it.

emichr:
> ```python
> s = hs.load(file, navigation_shape=(514,512))
> s = s.inav[:512,:]
> ```
> 
> Thus manually trimming the lineskip! @emichr let me know if you want to look into this example further.

Hi @sivborg, great that you tested this! Perhaps it is better to just modify the docstring with this as an example (and possibly making a note in the documentation as well) instead of including my suggested code changes? What do you think @ericpre?

ericpre:
A simple approach would be to specify the number of frames to skip, as mentioned in comment in https://github.com/hyperspy/rosettasciio/issues/270#issuecomment-2158650368. This would be more convenient for the user, rather having to remember/figure out where the add the additional frames.

I don't remember the details, but I would expect that it could be implemented without loading the frames to skip, which would be more efficient, particularly when loading lazily.

ericpre:
@emichr, have you been able to make the test files required to finish this pull request?

Issue #296, opened 2024-07-16
emd reader Velox ModuleNotFoundError 


dbdurham:
#### Describe the bug
Tried to read in a spectrum image Velox emd file (EDS) and got: 

```python
ModuleNotFoundError                       Traceback (most recent call last)
Cell In[4], line 1
----> 1 d = hs.load('example.emd')

File ~\.conda\envs\hyperspy\Lib\site-packages\hyperspy\io.py:537, in load(filenames, signal_type, stack, stack_axis, new_axis_name, lazy, convert_units, escape_square_brackets, stack_metadata, load_original_metadata, show_progressbar, **kwds)
    533         objects.append(signal)
    534 else:
    535     # No stack, so simply we load all signals in all files separately
    536     objects = [
--> 537         load_single_file(filename, lazy=lazy, **kwds) for filename in filenames
    538     ]
    540 if len(objects) == 1:
    541     objects = objects[0]

File ~\.conda\envs\hyperspy\Lib\site-packages\hyperspy\io.py:596, in load_single_file(filename, **kwds)
    590     raise ValueError(
    591         "`reader` should be one of None, str, " "or a custom file reader object"
    592     )
    594 try:
    595     # Try and load the file
--> 596     return load_with_reader(filename=filename, reader=reader, **kwds)
    598 except BaseException:
    599     _logger.error(
    600         "If this file format is supported, please "
    601         "report this error to the RosettaSciIO developers at "
    602         "https://github.com/hyperspy/rosettasciio/issues"
    603     )

File ~\.conda\envs\hyperspy\Lib\site-packages\hyperspy\io.py:618, in load_with_reader(filename, reader, signal_type, convert_units, load_original_metadata, **kwds)
    616 lazy = kwds.get("lazy", False)
    617 if isinstance(reader, dict):
--> 618     file_data_list = importlib.import_module(reader["api"]).file_reader(
    619         filename, **kwds
    620     )
    621 else:
    622     # We assume it is a module
    623     file_data_list = reader.file_reader(filename, **kwds)

File ~\.conda\envs\hyperspy\Lib\site-packages\rsciio\emd\_api.py:201, in file_reader(filename, lazy, dataset_path, stack_group, select_type, first_frame, last_frame, sum_frames, sum_EDS_detectors, rebin_energy, SI_dtype, load_SI_image_stack)
    199         raise IOError("The file is not a supported EMD file.")
    200 except Exception as e:
--> 201     raise e
    202 finally:
[...]

PR #305, opened 2024-08-26
EMD Velox: update error msg for missing sparse module

dbdurham:
### Description of the change
This is an update in response to Issue #296 "emd reader Velox ModuleNotFoundError." When attempting to load EDS Spectrum Image data from a Velox EMD file without the optional "sparse" dependency installed, a ModuleNotFoundError is raised which is confusing to some users. To clarify the issue, a try except statement was added around running the Velox EMD file reader and a new error message is printed. 

### Progress of the PR
- [x] add tests,
- [x] ready for review.

### DIFF
--- a/rsciio/emd/_api.py
+++ b/rsciio/emd/_api.py
@@ -173,7 +173,6 @@ def file_reader(
     try:
         if is_EMD_Velox(file):
             from ._emd_velox import FeiEMDReader
-
             _logger.debug("EMD file is a Velox variant.")
             emd_reader = FeiEMDReader(
                 lazy=lazy,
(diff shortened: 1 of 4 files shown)


ericpre (commented):
Thank you @dbdurham for the pull request, this looks good. Can you try to add a test in https://github.com/hyperspy/rosettasciio/blob/main/rsciio/tests/test_emd_velox.py to cover the code that you added? It will need to catch the error using `pytest.raises` but also update the `pytest.importorskip("sparse")` accordingly to enable running the test.
I am happy to do that if this helps.

Can you add a changelog entry? The instruction on how to do that are in https://github.com/hyperspy/rosettasciio/blob/main/upcoming_changes/README.rst.

dbdurham:
@ericpre sure I will try that, thanks for your suggestions as this is my
first time contributing to a large project like this. May take me a few
days to get to the tests and get it right if that's ok, as it's also my
first time working with pytest.

dbdurham:
@ericpre I will no longer be able to complete these changes, sorry for the
long delay

rsciio/emd/_api.py
```diff
@@ -167,6 +167,11 @@ def file_reader(
         spatial drift in the spectrum image by using the STEM images.
 
     %s
+
+    Raises
+    ------
+    ModuleNotFoundError
+        When reading spectrum image from Velox EMD file and the sparse is missing.
```
jlaehne, line 174: The sparse what? e.g. library as below?

```suggestion
        When reading spectrum image from Velox EMD file and the ``sparse`` library is missing.
```

PR #315, opened 2024-10-02
Reading/Writing DM5 files

CSSFrancis:
Curious of peoples thoughts on trying to support reading/writing of .dm5 files.  The format is pretty much identical to the  .dm4 and .dm3 file format except for split into a hierarchical data format. It doesn't seem too hard to write a (minimum) dm5 file that can be read.  This would be a good way for people to check/compare data and also good for interoperability. 

@jlaehne do you have any example datasets? That might be helpful.  I have some 4D STEM datasets but any other data type would be nice to help identify how the guide image should be created.  

### Description of the change
- [x] Read 4D STEM
- [x] Write 4D STEM
- [x] Read In situ TEM
- [x] Write In Situ TEM
- [x] Read EDS 
- [x] Write EDS
- [x] Read EELS
- [x] Write EELS
- [x] add tests,
- [ ] ready for review.

Any suggestions might be helpful here.  I think I mostly have a good handle on the file format but specifically I'm not sure how important the Meta Data tags are in identifying the "Type" of dataset.

### DIFF
--- a/rsciio/dm5/__init__.py
+++ b/rsciio/dm5/__init__.py
@@ -1,0 +1,8 @@
+from ._api import file_reader, file_writer
+
+__all__ = ["file_reader", "file_writer"]
+
+
+def __dir__():
+    return sorted(__all__)
+
--- a/rsciio/dm5/__init__.py
+++ b/rsciio/dm5/__init__.py
@@ -5,4 +5,3 @@ __all__ = ["file_reader", "file_writer"]
 
 def __dir__():
     return sorted(__all__)
-
--- a/rsciio/dm5/_api.py
+++ b/rsciio/dm5/_api.py
@@ -1,5 +1,20 @@
-from asyncore import write
-from importlib.metadata import metadata
+# -*- coding: utf-8 -*-
+# Copyright 2007-2023 The HyperSpy developers
+#
+# This file is part of RosettaSciIO.
+#
+# RosettaSciIO is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# RosettaSciIO is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with RosettaSciIO. If not, see <https://www.gnu.org/licenses/#GPL>.
 
 import h5py
 import dask.array as da
@@ -7,8 +22,6 @@ import numpy as np
 
 from rsciio._docstrings import FILENAME_DOC, LAZY_DOC, RETURNS_DOC
 from rsciio.hspy._api import HyperspyWriter
-from logging import getLogger
-_logger = getLogger(__name__)
 
 
 # for some reason this is slightly different in dm4???
@@ -547,8 +560,6 @@ def _group2dict(group, dictionary=None):
             value = value.astype("U")
             if value.dtype.str.endswith("U1"):
                 value = value.tolist()
-        else:
-            value = value
         dictionary[key] = value
 
     if not isinstance(group, h5py.Dataset):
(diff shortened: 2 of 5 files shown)


jlaehne:
> @jlaehne do you have any example datasets? That might be helpful. I have some 4D STEM datasets but any other data type would be nice to help identify how the guide image should be created.

Unfortunately our lab setup has some old interfaces so that we are stuck on a DM 2.72 version. So I can not directly produce sample data sets. If the latest free DM 3.x can write .dm5 files, which I have not looked in to, we could of course convert existing test data sets from .dm4 to .dm5 (otherwise I know one or two labs that might be able to produce such data sets for CL though).

If Gatan is already making the move to an hdf5 based data format that is openly documented, couldn't we try to convince them to contribute directly to the I/O plugin and maintain it?

ericpre:
I have a version of GMS that can save dm5 on my laptop, so I can easily make file when necessary.

ericpre:
https://www.gatan.com/dm5-documentation

CSSFrancis:
> If Gatan is already making the move to an hdf5 based data format that is openly documented, couldn't we try to convince them to contribute directly to the I/O plugin and maintain it?

If you would like to reach out to them that would be great!

CSSFrancis:
pre-commit.ci autofix

jlaehne:
> As far as ROI's go I'm not sure if those should be added by default

Well there have been discussions #22 and attempts concerning #1491 marker import for dm3/4 ... would still be a really nice feature also for the older file versions, but maybe optional and not as default.

CSSFrancis:
@jlaehne I might come back to this once I make some changes upstream in hyperspy.  Namely replacing the Navigator with a ROI and allowing multiple navigators.  That should make things more similar to DM and make the 1:1 mapping a little easier.

CSSFrancis:
@jlaehne The ROI's aren't terribly difficult to create although they appear to be unsupported with the free DM version?

jlaehne:
I currently don't have access to the free version of DM ... but @aidanc151 has looked into that a bit and has a local implementation for reading these ROIs - so maybe you could team up - also he could check on the free DM version.

ericpre:
I tried to make some dm5 file using GMS 3.6.0 and I had an issue with saving cropped data... so what I tried to save as dm4, then open in hyperspy and then saved in dm5 using hyperspy, which does work. These are the data:

When opening the data in GMS, it seems that the shape of the data is incorrect, hyperspy reads it correctly though! Some of the original metadata are lost, but I guess this is secondary at the moment.
[EELS_SI.zip](https://github.com/user-attachments/files/17445377/EELS_SI.zip)

The dm5 are made using:
```python
import hyperspy.api as hs

fname = "EELS_SI"

s = hs.load(f"{fname}.dm4")
s.save(f"{fname}.dm5")
s.T.save(f"{fname}_transposed.dm5")
```

When reading the transposed file in GMS, it seems that there is an encoding issue with the `µ`:
image

ericpre:
> Also let me know if there is interest in a DM3/DM4 writer. I should be able to reverse engineer those now that I understand the structure a little bit better.

From https://github.com/hyperspy/ImportRPL/issues/2#issuecomment-1189977590 and having a quick look, https://github.com/mfm24/DM3Utils may be a good reference to implement writing dm4 file.

CSSFrancis:
> I tried to make some dm5 file using GMS 3.6.0 and I had an issue with saving cropped data... so what I tried to save as dm4, then open in hyperspy and then saved in dm5 using hyperspy, which does work. These are the data:
> 
> When opening the data in GMS, it seems that the shape of the data is incorrect, hyperspy reads it correctly though! Some of the original metadata are lost, but I guess this is secondary at the moment. Below are some cropped data: [EELS_SI.zip](https://github.com/user-attachments/files/17445377/EELS_SI.zip)

This is a good sign right? :)  Saving the metadata is bit more of a challenge considering that both Hyperspy and DM both have pretty much arbitrary metadata so mapping 1:1 is pretty much impossible.  

I just dump the hyperspy metadata into a branch and vice-versa although that isn't the best for the case of going from DM --> Hyperspy --> DM :) 

> 
> The dm5 are made using:
> 
> ```python
> import hyperspy.api as hs
> 
> fname = "EELS_SI"
> 
> s = hs.load(f"{fname}.dm4")
> s.save(f"{fname}.dm5")
> s.T.save(f"{fname}_transposed.dm5")
> ```

Issue #316, opened 2024-10-03
Failed assert for lazily-loaded tiff stack


realitychemist:
#### Describe the bug
I have simulated 4D-STEM data (output from the Dr Probe software) that's stored as a 3D TIFF stack with a companion JSON file that contains the metadata (e.g. the correct shape for the 4D axes). I am trying to load the TIFF file, reshape it according to the information in the JSON file, and save the reshaped data in EMD (NCEM) format (which is the native file format for py4DSTEM).

The 3D TIFF stack is ~180 GB in size and will not fit in RAM, so I am loading it lazily via `tiff.file_reader(tiff_fname, lazy=True)`. This appears to work at first: I can (lazily) create a reshaped `dask` array based on the JSON file (loaded separately with the `json` library) and I can construct the `axes` and `metadata` objects with no issue (all necessary information is in the JSON file).

When I go to actually write out the new EMD file (via `emd.file_writer(emd_fname, signal)`) an `AssertionError: assert self._fh is not None` is raised. Here's the stack trace (with paths trimmed up to "site-packages" for readability):

> Traceback (most recent call last):
  File "\<ipython-input-7-6d09027a0035\>", line 2, in <module>
    emd.file_writer(emd_fname, signal)
  File "...\\rsciio\emd\_api.py", line 230, in file_writer
    EMD_NCEM().write_file(filename, signal, chunks=chunks, **kwds)
  File "...\\rsciio\emd\_emd_ncem.py", line 474, in write_file
    self._write_signal_to_group(signal_group, signal, **kwargs)
  File "...\\rsciio\emd\_emd_ncem.py", line 500, in _write_signal_to_group
    dataset.create_dataset(
  File "...\\h5py\_hl\group.py", line 183, in create_dataset
    dsid = dataset.make_new_dset(group, shape, dtype, data, name, **kwds)
  File "...\\h5py\_hl\dataset.py", line 46, in make_new_dset
    data = array_for_new_object(data, specified_dtype=dtype)
  File "...\\h5py\_hl\base.py", line 118, in array_for_new_object
    data = np.asarray(data, order="C", dtype=as_dtype)
  File "...\\dask\array\core.py", line 1693, in __array__
    x = self.compute()
  File "...\\dask\base.py", line 375, in compute
    (result,) = compute(self, traverse=False, **kwargs)
  File "...\\dask\base.py", line 661, in compute
    results = schedule(dsk, keys, **kwargs)
[...]

PR #317, opened 2024-10-05
Fix loading tiff file lazily

ericpre:
Fix for https://github.com/hyperspy/rosettasciio/issues/316: using the context manager to load tiff files was closing the file. Some tiff were still loading lazily (for example the file used to test lazy loading in the test suite before this PR) because a different code path was being used.

### Progress of the PR
- [x] Fix loading tifffile lazily,
- [x] add support for tiff file in `get_file_handle` 
- [x] add tests,
- [x] ready for review.

### DIFF
--- a/rsciio/utils/tools.py
+++ b/rsciio/utils/tools.py
@@ -503,27 +503,59 @@ def ensure_unicode(stuff, encoding="utf8", encoding2="latin-1"):
 
 
 def get_file_handle(data, warn=True):
-    """Return file handle of a dask array when possible; currently only hdf5 file are
-    supported.
     """
-    arrkey = None
+    Return file handle of a dask array when possible.
+    Currently only hdf5 and tiff file are supported.
+
+    Parameters
+    ----------
+    data : dask.array.Array
+        The dask array from which the file handle
+        will be retrieved.
+    warn : bool
+        Whether to warn or not when the file handle
+        can't be retrieved. Default is True.
+
+    Returns
+    -------
+    File handle or None
+        The file handle of the file when possible.
+    """
+    arrkey_hdf5 = None
+    arrkey_tifffile = None
     for key in data.dask.keys():
         # The if statement with both "array-original" and "original-array"
         # is due to dask changing the name of this key. After dask-2022.1.1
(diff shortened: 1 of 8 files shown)


rsciio/utils/tools.py
```diff
@@ -503,27 +503,59 @@ def ensure_unicode(stuff, encoding="utf8", encoding2="latin-1"):
 
 
 def get_file_handle(data, warn=True):
-    """Return file handle of a dask array when possible; currently only hdf5 file are
-    supported.
     """
-    arrkey = None
+    Return file handle of a dask array when possible.
```
CSSFrancis, line 507: How does this work with memmaped files? 

We should make a new issue to maybe make this a bit more routine.

CSSFrancis: What if we put the original file into the metadata or added it to the signal dictionary.  Would that be slightly better as we could just track the initial object.

CSSFrancis (approved):
This is a good change, although what happens in the case of:


```python
s = hs.load("file.tiff", lazy=True)
summed = s.sum()
summed.compute() # This shouldn't close the file correct?
```


```python
s = hs.load("file.tiff", lazy=True)
summed = s.sum()
s.compute()
summed # This should still be lazy with the file not closed, correct?
```

ericpre:
The API to close file is in [`compute(close_file=True)`](https://hyperspy.org/hyperspy-doc/current/reference/base_classes/signal/LazySignal.html#hyperspy._signals.lazy.LazySignal.compute) with the default set to False.

> This is a good change, although what happens in the case of:
> 
> ```python
> s = hs.load("file.tiff", lazy=True)
> summed = s.sum()
> summed.compute() # This shouldn't close the file correct?
> ```
Yes, this will not close the file.

> ```python
> s = hs.load("file.tiff", lazy=True)
> summed = s.sum()
> s.compute()
> summed # This should still be lazy with the file not closed, correct?
> ```
Same as above, this will not close the file.

rsciio/utils/tools.py
ericpre: Agree, we need an API to do that.

Issue #345, opened 2024-12-12
CI Failing with Dask 2024.12.0


CSSFrancis:
#### Describe the bug

Test passes with dask - 2024.11.2 and fails with dask 2024-12.0

#### To Reproduce

```python
rsciio/tests/test_blockfile.py:438 (test_load_readonly)
def test_load_readonly():
        s = hs.load(FILE2, lazy=True)
        k = next(
            filter(
                # The or statement with both "array-original" and "original-array"
                # is due to dask changing the name of this key. After dask-2022.1.1
                # the key is "original-array", before it is "array-original"
                lambda x: isinstance(x, str)
                and (x.startswith("original-array") or x.startswith("array-original")),
                s.data.dask.keys(),
            )
        )
        mm = s.data.dask[k]
        assert isinstance(mm, np.memmap)
>       assert not mm.flags["WRITEABLE"]
E       assert not True

test_blockfile.py:453: AssertionError
```

#### Additional context
Will look upstream to determine what has changed.

CSSFrancis:
Failutre from Commit https://github.com/dask/dask/commit/511b8af29b3b1f4b60cb56dff7f722ffc01179f9 

https://github.com/dask/dask/pull/11524

CSSFrancis:
@ericpre do you have any idea if this loads thinggs into memory? It seems like it doesn't but the change in the array flags is slightly concering. 

```
import numpy as np
# Create a memmap array backed by a file
filename = 'my_array.dat'
dtype = np.float32
shape = (100, 100)

# Create the memmap
mmap_array = np.memmap(filename, dtype=dtype, mode='r', shape=shape)

m2 = mmap_array.copy() 

m2 # mmap array still
```

ericpre:
Sorry no idea...

CSSFrancis:
Okay from my testing this is different but shouldn't be detrimental.  You won't be able to write to the original array, but that is already the prefered behavior.  There isn't a huge overhead for the copy so we can probably just remove the check.

ericpre:
Sounds good, is there actually a simple way to test the behaviour (not possible to write?), maybe check that it raises an error?

I am wondering, if this is still the case that the data can't be written? Could it be related to the fact that previously, dask array were not mutable but this has changed since the blockfile code has been written?

CSSFrancis:
> Sounds good, is there actually a simple way to test the behaviour (not possible to write?), maybe check that it raises an error?

So if we take this little bit of code:

```python
[...]

PR #346, opened 2024-12-13
Testing: Comment out check

CSSFrancis:
### Description of the change
A few sentences and/or a bulleted list to describe and motivate the change:
- Remove check that memmap file isn't "writable"
- See #345  for more details. 
-
### Progress of the PR
- [ ] ready for review.


ericpre (approved):
Should we merge as it is and revisit later or is there anything else you would like to try?

CSSFrancis:
I think this is all we can really do here. I don't foresee this as an issue. I think it's just different not necessarily wrong.

PR #347, opened 2024-12-16
Fix hamamatsu tiff reader if x axis not calibrated

jlaehne:
### Description of the change
Fixes reading error when Hamamatsu tiff file contains `ScalingXScalingFile="Other"` and not an address in the file (not calibrated axis).

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
from rsciio.tiff import file_reader
file_reader("hamamatsu.tiff")
# Your new feature...
```


jlaehne:
pre-commit.ci autofix

CSSFrancis (approved):
Looks like a pretty simple bug fix!

rsciio/tiff/_api.py
```diff
@@ -696,7 +696,11 @@ def _get_hamamatsu_streak_description(tiff, op):
     scaling = dict_meta["Scaling"]
 
     # Address in file where the X axis is saved
-    x_scale_address = int(re.findall(r"\d+", scaling["ScalingXScalingFile"])[0])
+    # If x axis is "Other" it just loads the axis as pixels
+    if scaling["ScalingXScalingFile"].startswith("Other"):
```
CSSFrancis, line 700: How does this fail currentlt. If for example the "ScalingXScalingFile" were set to "custom" would this fail again?

Maybe consider something like a try --> except in the else statement so that the data is still loaded but it throws a warning saying the scale could not be properly loaded.

rsciio/tiff/_api.py
```diff
@@ -696,7 +696,11 @@ def _get_hamamatsu_streak_description(tiff, op):
     scaling = dict_meta["Scaling"]
 
     # Address in file where the X axis is saved
-    x_scale_address = int(re.findall(r"\d+", scaling["ScalingXScalingFile"])[0])
+    # If x axis is "Other" it just loads the axis as pixels
```
ericpre, line 699: Would it make sense to add comment explaining in what situation is the x axis "Other"?

ericpre (approved):

rsciio/tiff/_api.py
jlaehne: I amended the comment

rsciio/tiff/_api.py
jlaehne: In principle yes, but normally it contains an address in the file where the axis is saved - if that is not the case it is set to "Other". That is not documented: http://www-bl20.spring8.or.jp/detectors/manual/HelpHiPic.pdf (Appendix D, E). However, I do not expect hamamatsu to have different keywords that can go there in place of an address, so I think it might even be better to get an error if there is yet another situation and not do a try and except for situations that are not documented.

ericpre: Generally speaking, I am on fence regarding @CSSFrancis's suggestion:
- in principle this is sensible, if there is a warning to suggest to report this issue to avoid a silent failure (here calibration not loading).
- in practise, raising an error motivates user to report error, while a warning will be ignored usually (and a source of noise or polluting log, notebook, etc. for some users) and it means that there is a risk that it doesn't get fixed!

PR #356, opened 2025-01-16
Custom Scan .mrc

CSSFrancis:
### Description of the change

Direct Electron allows for custom scan patterns/ non-linear distributed or serpentine scan patterns to be loaded.  This is a nice feature, but sometimes it can be difficult to account for when loading data.  Ultimately, the data is still locked to a grid, but no every pixel has to be scanned and the data is often not in the correct order. 

Along with each .mrc file a scan_coordinates.csv file is also saved which is simply two columns with x, y scan positions. This file is likely to be updated in future with additional columns (time stamp, in situ information etc) but for now it should be relatively static. 

Loading this type of data is a little tricky, but the core concept is that a fully dense array is created. If a scan position isn't scanned, it is set to all zeros and should be easy to ignore, mask etc. and it should compress down to almost nothing.  Otherwise I could set those missing arrays to all nan but that would require casting from int to float which I would rather not do. 

### Progress of the PR
- [x] Add loading custom scans
- [x] Zeros where no scan is
- [x] add tests,
- [ ] ready for review.

```python
s = hs.load("test_scan.mrc", scan_file="scan_coordinates.csv")
```

### DIFF
--- a/rsciio/mrc/_api.py
+++ b/rsciio/mrc/_api.py
@@ -490,7 +490,6 @@ def file_reader(
     if not lazy:
         data = data.compute()
 
-
     original_metadata["std_header"] = sarray2dict(std_header)
 
     # Convert bytes to unicode


CSSFrancis:
pre-commit.ci autofix

CSSFrancis:
pre-commit.ci autofix

ericpre (approved):
This looks good, it just needs to fix the conflicts after https://github.com/hyperspy/rosettasciio/pull/402 was merged.

CSSFrancis:
pre-commit.ci autofix

PR #375, opened 2025-03-07
Add file handle API

### LINKED ISSUE #331: Add API for file handle

ericpre:
Closes #331.

### Progress of the PR
- [x] Add `file_handle` API,
- [x] implement `file_handle` API for the following format,
  - [x] hspy
  - [x] emd
  - [x] nexus
  - [x] usid    
  - [x] tiff 
- [x] add tests,
- [x] ready for review.

### DIFF
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -35,7 +35,7 @@ The **dictionary** contains the following fields:
 * ``'metadata'`` -- dictionary containing the parsed metadata
 * ``'original_metadata'`` -- dictionary containing the full metadata tree from the
   input file
-* ``'file_handle'`` -- (compatible format only: :ref:`emd <emd-format>`, :ref:`hspy <hspy-format>`, :ref:`NeXus <nexus-format>`, :ref:`tiff <tiff-format>`, :ref:`USID <usid-format>`) the file handle when ``lazy`` is ``True`` otherwise ``None``
+* ``'file_handle'`` -- (compatible formats restricted to: :ref:`emd <emd-format>`, :ref:`hspy <hspy-format>`, :ref:`NeXus <nexus-format>`, :ref:`tiff <tiff-format>`, :ref:`USID <usid-format>`) the file handle when ``lazy`` is ``True`` otherwise ``None``
 
 Interfacing the reader from one of the IO plugins:
 
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -35,6 +35,7 @@ The **dictionary** contains the following fields:
 * ``'metadata'`` -- dictionary containing the parsed metadata
 * ``'original_metadata'`` -- dictionary containing the full metadata tree from the
   input file
+* ``'file_handle'`` -- (compatible format only) the file handle when ``lazy`` is ``True`` otherwise ``None``
 
 Interfacing the reader from one of the IO plugins:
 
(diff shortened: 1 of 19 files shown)


doc/api/index.rst
```diff
@@ -35,6 +35,7 @@ The **dictionary** contains the following fields:
 * ``'metadata'`` -- dictionary containing the parsed metadata
 * ``'original_metadata'`` -- dictionary containing the full metadata tree from the
   input file
+* ``'file_handle'`` -- (compatible format only: :ref:`emd <emd-format>`, :ref:`hspy <hspy-format>`, :ref:`NeXus <nexus-format>`, :ref:`tiff <tiff-format>`, :ref:`USID <usid-format>`) the file handle when ``lazy`` is ``True`` otherwise ``None``
```
jlaehne, line 38:
```suggestion
* ``'file_handle'`` -- (compatible formats restricted to: :ref:`emd <emd-format>`, :ref:`hspy <hspy-format>`, :ref:`NeXus <nexus-format>`, :ref:`tiff <tiff-format>`, :ref:`USID <usid-format>`) the file handle when ``lazy`` is ``True`` otherwise ``None``
```

rsciio/_docstrings.py
```diff
@@ -135,6 +135,7 @@
           containing the full axes vector
         - 'metadata' – dictionary containing the parsed metadata
         - 'original_metadata' – dictionary containing the full metadata tree from the input file
+        - 'file_handle' - (compatible format only) the file handle when ``lazy`` is ``True`` otherwise ``None``
```
jlaehne, line 138:
```suggestion
        - 'file_handle' - (compatible formats only) the file handle when ``lazy`` is ``True`` otherwise ``None``
```

jlaehne (approved):

ericpre:
@CSSFrancis, what do you think about this API?

CSSFrancis:
@ericpre I really like this change and think it was a long time coming.  That being said I think we need to think a little bit about more how we are handling files and what the relationship between hyperspy --> dask --> underlying file.  More and more it's seeming like `dask` not being "file aware" is a bit of a problem. 

More than likely what we need is a abstract class which is something like https://docs.xarray.dev/en/latest/generated/xarray.backends.CachingFileManager.html which would allow File Objects to be passed from one process to the next without pickling.  This is how `xarray` handles things like distributed hdf5 processing which would be wonderful :).  

To maybe take it a step further, I would like to build a bit more context support for things like rechunking.  I think this would involve something like the optimized slicing that `b2h5py` does:

https://github.com/Blosc/b2h5py/blob/317f4a0c31ac61623bf49861f3f5c6ebce46f633/b2h5py/blosc2.py#L248

Mostly it would just be optimized slicing for things like virtual images, plotting etc.

ericpre:
@CSSFrancis, thank you, this sounds good.

Regarding this PR, would it be better to put it on hold until a file manager is implemented to return the file manager object instead of the file handle? Even if they may have similar API, there are _not_ the same and therefore can't be swapped without breaking the API?

PR #379, opened 2025-03-16
Build free threaded python

ericpre:
Build free threaded python 3.13 to make easier to install in free threaded environment.
There are a few workaround because h5py doesn't have a free threaded wheel yet and a pre-release of cython needs to be used.

### Progress of the PR
- [x] Setup free threaded on CI following https://py-free-threading.github.io/ci guidance,
- [x] declare bruker c extension as compatible with free threading to avoid enabling the GIL when loading extension, 
- [x] add to release workflow 
- [n/a] add tests,
- [x] ready for review.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,6 +39,7 @@ jobs:
         uses: pypa/cibuildwheel@v2.23
         env:
           CIBW_ARCHS: ${{ matrix.CIBW_ARCHS }}
+          CIBW_BUILD_FRONTEND: 'pip; args: --pre --extra-index-url "https://pypi.anaconda.org/scientific-python-nightly-wheels/simple"'
 
       - name: List wheels
         run: |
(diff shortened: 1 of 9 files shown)


ericpre:
@CSSFrancis, @jlaehne, I suggest to include this PR in the 0.8 release to be able to start testing freethreaded python and the GIL doesn't get enable when loading bcf files.

CSSFrancis (approved):
@ericpre I looked over this and it looks good.  From the cython side of things this still requires that we rewrite the code to be multithreaded? Correct? I haven't look into the `unbcf_fast.pyx` code very much.  

I haven't been following the free threaded updates in python 3.13 very closely. Isn't the GIL released for cython code already?

ericpre:
The main thing when moving to freethreaded python is to consider whether the code is thread-safe or not. With the GIL, we don't have to worry about this because there is only a single thread that is running at any one time and the threads take turns.... with freethreaded, the GIL is disabled and the main issue to consider is that different threads can change the state of an object at the same time.

Regarding the bruker bcf file, my understanding of the cython code is that it runs serially (when loading the EDS stream) and there is no multithreading in its current implementation.

Isn't the GIL released for cython code already?
No, but it can be specified: https://cython.readthedocs.io/en/latest/src/userguide/nogil.html

PR #386, opened 2025-04-05
Set correct signal type of Hamamatsu tiff images, if lumispy is installed

jlaehne:
Set `signal_type` of Hamamatsu streak images loaded from `tiff` format to `TransientSpectrum` (`LumiTransientSpectrum` class), if lumispy is installed.

Add tests to tiff and hamamatsu readers that run when lumispy is available.

### Progress of the PR
- [x] add tests,
- [x] ready for review.


rsciio/hamamatsu/_api.py
```diff
@@ -359,7 +359,7 @@ def _map_signal_md(self):
             )
             signal["signal_type"] = ""
         else:
-            signal["signal_type"] = "LumiTransientSpectrum"  # pragma: no cover
+            signal["signal_type"] = "LumiTransientSpectrum"
```
ericpre, line 362: Would it be worth adding a build on GitHub CI that install extensions (lumispy, pyxem, etc.) to test this kind of thing?

jlaehne: Possibly. That is what the integration tests on the `hyperspy-extensions-list` repro are doing - but then it is not noticed as directly if something goes wrong. I can also add back the `pragma: no cover` and we leave it to the integration tests to run this part?

ericpre: > That is what the integration tests on the `hyperspy-extensions-list` repro are doing - but then it is not noticed as directly if something goes wrong.

Indeed, but this can be tedious when fixing it or figure out what needs to be fixed. We could use the same approach as in hyperspy with a separate workflow that runs on demand (https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/.github/workflows/tests_extension.yml) but we trend to forget about it and also it doesn't help with the coverage.

jlaehne: Adding such tests in #388

ericpre: Regarding the coverage, should we add back `# pragma: no cover`, to silent coverage warning and add a comment that this is tested in the integration test workflow but the coverage is not reported?

jlaehne: done

ericpre (approved):

PR #387, opened 2025-04-05
Fix hamamatsu img reader if x axis not calibrated

jlaehne:
Fixes reading error when Hamamatsu img file contains ScalingXScalingFile="Other" and not an address in the file (not calibrated axis). In this case, the size was not set correctly and data array could not be rescaled.

### Progress of the PR
- [x] add tests,
- [ ] ready for review.

```python
from rsciio.hamamatsu import file_reader
file_reader("xaxis_other.img")
```


rsciio/tests/test_hamamatsu.py
```diff
@@ -399,3 +400,30 @@ def test_metadata(self):
...
+    @classmethod
+    def setup_class(cls):
+        cls.s = hs.load(testfile_xaxisother_path, reader="Hamamatsu")
+
+    @classmethod
+    def teardown_class(cls):
+        del cls.s
+        gc.collect()
+
+    def test_xaxisother(self):
```
ericpre, line 415: Since there is only test in this class, would it be more simple to use a function instead of a class?

rsciio/tests/test_hamamatsu.py
```diff
@@ -399,3 +400,30 @@ def test_metadata(self):
...
+        # Test case where ScalingXScalingFile="Other"
+        assert self.s.axes_manager.signal_shape == (672, 508)
+        assert self.s.axes_manager.navigation_shape == ()
+        assert self.s.data.shape == (508, 672)
+        assert self.s.axes_manager[0].units == "px"
+        assert self.s.axes_manager[1].units == "px"
```
ericpre, line 421: Check the name of the axis?

rsciio/tests/test_hamamatsu.py
jlaehne: As there is a class per file used for the tests of this reader, I wanted to stick to that pattern.

ericpre (approved):

PR #388, opened 2025-04-06
Add integration tests

jlaehne:
Add test run with hyperspy and its extensions installed. Run rosettasciio, hyperspy and extension tests.

### Progress of the PR
- [x] add tests,
- [x] ready for review.


ericpre:
Just ran the workflow in hyperspy: https://github.com/hyperspy/hyperspy/actions/runs/14291193279/job/40052635962 and there are a few test suite that fails and from a quick look, it seems to be related to an issues in the test suite of the extension.

For now, a pragmatic approach could be to skip these test suite, until this is fixed - running rosettasciio test suite with the extension installed is already an improvement and useful.

At some point, it would be good to see if reusable workflow can be used, otherwise, this can be end up being a maintenance burden or not working most of the time!

jlaehne:
Indeed, it is probably enough to run the Rosetta tests with extensions installed here and leave further integration tests to HyperSpy.

.github/workflows/tests_hs-extension.yml
```diff
@@ -0,0 +1,80 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        EXTENSION_VERSION: ['release', 'dev']
+
+    env:
+      MPLBACKEND: agg
+      EXTENSION: atomap etspy exspy holospy kikuchipy lumispy pyxem
+      TEST_DEPS: pytest pytest-xdist pytest-rerunfailures pytest-instafail
```
ericpre, line 24: From https://github.com/hyperspy/rosettasciio/blob/c53293a1208e58575c2820d135adb779dd9b3536/pyproject.toml#L102-L109

```suggestion
      TEST_DEPS: pytest pytest-xdist pytest-rerunfailures pytest-instafail filelock pooch
```

ericpre (commented):
Should it be documented somewhere, pull request template or contributor guide?

PR #392, opened 2025-04-17
Remove optional dependency (`blosc`) of `mrcz`

ericpre:
`blosc2` is now replacing `blosc` and wheels are not available starting from python 3.13, which causes the installation of `blosc` on CI to fail with python 3.13.
Once `mrcz` supports `blosc2`, it may be worth considering adding it back as a convenience.

### Progress of the PR
- [x] Remove `python-mrcz` dependency, which was added as a convenience,
- [n/a] add tests,
- [ ] ready for review.


CSSFrancis:
@ericpre Is MRCZ using blosc2 something worth supporting?  I've really liked the blosc2 libary when I looked into it but unfortunately it doesn't seem to be widely used in practice.  Many of the cryo-em softwares don't even support mrcz even though it is quite a bit nicer than something like tiff-lzw.   I've considered writing a MRCZ implementation with blosc2 but I'd need to make sure that it would actually be used/useful rather than just another file format.

CSSFrancis (approved):
Also this looks good to me!

ericpre:
> @ericpre Is MRCZ using blosc2 something worth supporting? I've really liked the blosc2 libary when I looked into it but unfortunately it doesn't seem to be widely used in practice. Many of the cryo-em softwares don't even support mrcz even though it is quite a bit nicer than something like tiff-lzw. I've considered writing a MRCZ implementation with blosc2 but I'd need to make sure that it would actually be used/useful rather than just another file format.

I have no idea how widely used mrcz is, looking at the download per month on prefix.dev: https://prefix.dev/channels/conda-forge/packages/mrcz, doesn't seem to be that much.
I agree that the improvement on large data performance should be beneficial but are there any acquisition software that write to this format? Another thing is that I am not sure if https://github.com/em-MRCZ/python-mrcz (and the conda-forge recipe) is maintained anymore. I tried to submit pull request to fix deprecation but I don't get any reply. An alternative to this would be to implement the `mrcz` support without depending on the `mrcz` library, as there is already a `mrc` reader, maybe it wouldn't be that much work to do and maintain?

@robbmcleod, would you be able to comment on these points?

Issue #377, opened 2025-03-11
Hspy distributed file reader


CSSFrancis:
#### Describe the functionality you would like to see.
It should be possible to load hdf5 files using dask distributed.  I think the major thing we need to do is add in the ability to pickle h5py objects.  

We should be able to drop in https://github.com/DaanVanVugt/h5pickle

or write our own code for it --> https://github.com/pydata/xarray/issues/4242

PR #398, opened 2025-04-24
Arina reader

smribet:
### Description of the change
Adding a reader for Arina data

### Progress of the PR
- [x] Adding arina reader
- [x] update docstring
- [x] update user guide
- [x] add tests
- [ ] ready for review

```python
from rsciio.arina import file_reader
file_reader("your_arina_file.h5")

### CODE REVIEW

File: rsciio/arina/specifications.yaml [hunk 1]
```diff
@@ -0,0 +1,4 @@
+name: arina
+description: Reader for arina 4D-STEM datasets
+extensions: ['.h5']
+non_uniform_axis: false
```
ericpre, line 4: Some keys are missing and this is what make the test suite fails, see the full list in https://hyperspy.org/rosettasciio/contributing.html#defining-new-rosettasciio-plugins.

rsciio/arina/_api.py
```diff
@@ -0,0 +1,245 @@
...
+import hdf5plugin  # noqa F401
+import numpy as np
+
+_logger = logging.getLogger(__name__)
+
+
+def file_reader(
+    filename,
+    lazy=False,
+    scan_width=None,
```
ericpre, line 32: We try to enforce consistent parameters name between reader and the `scan_width` would be similar to `navigation_shape`. If you want to define only one dimension, maybe something like `navigation_shape=(100, "auto")` can be used?

rsciio/arina/_api.py
```diff
@@ -0,0 +1,245 @@
...
+    binfactor=1,
+    dataset_dtype=None,
```
ericpre, line 34: Along the same line as the comment above, the velox and jeol reader have `rebin_energy` and `SI_dtype` parameters name. Their meaning is different to the parameters but would using `rebin` and `dtype` as these be closer to the existing parameter names?

ericpre (commented):
Looks like a good start! 😃 

One of the comments should help to fix the test suite - there will still some failure left because of https://github.com/hyperspy/rosettasciio/pull/397.

smribet:
Thanks for the helpful suggestions @ericpre! 

What is your preference for adding dependencies? The arina reader needs both hdf5plugin and the h5py, which is a little different than the existing h5py option.  I think this is also causing some of the errors for the pytests.

rsciio/arina/_api.py
```diff
@@ -0,0 +1,253 @@
...
+    rebin_diffraction=1,
+    dtype=None,
+    flatfield=None,
+):
+    """Read arina 4D-STEM datasets.
+
+    Parameters
+    ----------
+    navigation_shape : int, optional
+        Real space dimension of scan. If None, assumes a square acquisition.
```
ericpre, line 42: Actually, passing an integer (in addition to the tuple) may be more simple than specifying a tuple with "auto", not sure if "auto" is necessary?

The parameter `navigation_shape` should have the same signature as for other reader and this is also we reuse the docstring:
https://github.com/hyperspy/rosettasciio/blob/be2a6bd4bf26c631fbe564191247b22bb4b0d581/rsciio/mrc/_api.py#L553

https://github.com/hyperspy/rosettasciio/blob/be2a6bd4bf26c631fbe564191247b22bb4b0d581/rsciio/_docstrings.py#L79-L83


However, other reader will need to be updated to support it and it would be better to do it in a follow up PR - ideally before it is released.

ericpre (commented):
> What is your preference for adding dependencies? The arina reader needs both hdf5plugin and the h5py, which is a little different than the existing h5py option. I think this is also causing some of the errors for the pytests.

Adding the arina "extra", as you did, sounds good.

rsciio/arina/_api.py
smribet: Yes in this case since the total number of diffraction patterns is known, I tried to leave it flexible. I hope this matches better with the other readers!

rsciio/arina/specifications.yaml
```diff
@@ -0,0 +1,7 @@
+name: arina
```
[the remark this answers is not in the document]
smribet: Fantastic, thanks for catching this!

ericpre:
To fix some of the failure in the test suite, this branch will need to be rebased on the [main](https://github.com/hyperspy/rosettasciio/tree/main) branch. On the topic of rebasing, are you planning to use the file `rsciio/tests/data/arina/MEA_ptycho_15_data_000001.h5`? If not, it would be good to remove it from the git history to avoid unnecessarily increasing the repository size. I am happy to do it if this helps.

smribet:
@ericpre  if you wanted to do the rebasing that would be great! Unfortunately all 3 files are needed for the arina reader. Each dataset has a master file, at least one data file (although in bigger scans I can end up with more than 100 files), and then sometime an additional un-numbered file with metadata.

ericpre:
I rebased (and force-push to your branch) and pushed a commit of fix the tests that were failing because of the added optional dependencies.
You will need to update your branch locally from your branch on GitHub before pushing any more commits.

smribet:
This is fantastic! Thank you so much @ericpre! I replaced the test file to a much smaller one.

ericpre (commented):
@smribet, I rebase another time to fix the test suite and remove the large file from the history.

This looks good and a section needs to be added in the user guide before merging. You can see an example of what needs to be done in this PR: https://github.com/hyperspy/rosettasciio/pull/174/files#diff-ebf54ae884ed49064cfaec6f3ee2d8e37fff65f984c8f121cf7b3fc236f7d524 

There are two things that will be good to do in a follow PR:
- consolidate the `navigation_shape` argument with other reader. This PR extend the current syntax with using `"auto"` and it would be good to do that in another PR.
- add lazy support.

CSSFrancis:
One other comment.  I think that Dectris does all of their compression on GPU.  This is nice for the speed of compression but requires that you have very small chunks as a trade off.  As a result loading data on CPU might be much slower than it should be, mostly due to some limitations with the HDF5 library.  It might be worth benchmarking and trying to improve how fast we can load this data.  That or implement a loading scheme that uses a GPU to decompress the data if available. 

I think that #377 should help with this but I need to spend a good chunk of time determining exactly how to speed that up.

smribet:
These are great suggestions, thanks! I would be interested in these features in the future.  I tried to copy the other readers and add the relevant arina information to the docs. I wasn't sure about a few things, so feel free to send any more feedback! By the way -- is there a way to build the docs locally to check them before pushing?

rsciio/arina/_api.py
```diff
@@ -0,0 +1,272 @@
...
+        A tuple can be specified as (x_scan_dimension, y_scan_dimension), (x_scan_dimension, "auto"),
+        or ("auto", y_scan_dimension). With "auto" the length is inferred from the number of
+        diffraction patterns.  If only an integer is passed, it assumed to be the x_scan_dimension.
+    rebin_diffraction : int, default=1
+        Diffraction space binning factor for bin-on-load.
+    dtype : float, optional
+        Datatype for dataset.
+    flatfield : numpy.ndarray, optional
+        Flatfield for correction factors, converts data to float.
+    %s
```
CSSFrancis, line 63: PR02: Unknown parameters {'-------', 'Returns', 'list of dict'}
  PR04: Parameter "Returns" has no type
  PR07: Parameter "Returns" has no description
  PR04: Parameter "-------" has no type
  PR07: Parameter "-------" has no description
  PR04: Parameter "list of dict" has no type
  RT01: No Returns section found  --> You need an extra return to get numpydoc to know this is a different section
```suggestion

    %s
```

CSSFrancis (commented):
@smribet Also realizing some of these errors aren't the most helpful :) I think this is why things are failing.

ericpre (approved):

PR #400, opened 2025-05-19
Fix memory usage in lazy loading of mib file

ericpre:
With recent version of dask the non distributed lazy loading loads the whole dataset in memory. This implementation is not supposed to work anymore, (see for example: https://github.com/dask/dask/issues/11850) therefore, it is removed in this PR and only the distributed memory mapping is kept.

### Progress of the PR
- [x] Remove non-distributed memory mapping implementation in quantum detector reader,
- [x] fix passing `chunks` - same as #395.
- [x] update tests,
- [x] ready for review.

### DIFF
--- a/rsciio/quantumdetector/_api.py
+++ b/rsciio/quantumdetector/_api.py
@@ -326,13 +326,14 @@ def load_mib_data(
                 data = data.compute()
             data = data.squeeze()
     elif isinstance(path, bytes):
-        data = np.frombuffer(
+        buffer = np.frombuffer(
             path,
             dtype=merlin_frame_dtype,
             count=mib_prop.xy,
             offset=mib_prop.offset,
-        )["data"]
-        data = data.reshape(navigation_shape + mib_prop.merlin_size).squeeze()
+        ).reshape(navigation_shape)
+        headers = buffer["header"]
+        data = buffer["data"]
 
     else:  # pragma: no cover
         raise TypeError("`path` must be a str or a buffer.")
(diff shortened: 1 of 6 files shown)


ericpre:
@jlaehne, @CSSFrancis, could you please review this PR?

Once this is merged, I would suggest to make a release to support recent changes in tifffile and dask.

CSSFrancis:
@ericpre This should be all memmaped files right?  Lets just remove the distributed option and just use it by default. It works better, seems to be faster and we aren't constantly fighting with dask then :)

ericpre:
Yes, this is what PR does: the distributed memory mapping implementation is now used and the non-distributed memory mapping is removed. It also simplify the code.
Non-lazy loading, also used this lazy implementation, but this was already the case before this PR.

ericpre:
ok, should we do this in a separate PR?

CSSFrancis (approved):
Looks simple enough. Did you test what happens when someone has `hs.load("file.mib", distributed =True)` just to make sure that it still works.

rsciio/quantumdetector/_api.py
```diff
@@ -607,7 +581,6 @@ def file_reader(
         navigation_shape=navigation_shape,
         first_frame=first_frame,
         last_frame=last_frame,
-        distributed=distributed,
```
CSSFrancis, line 610: I think you might need to have a **kwargs argument so that having distributed=True doesn't cause an error.  We probably don't want to break a bunch of peoples code?

CSSFrancis:
> ok, should we do this in a separate PR?

Probably, This looks good btw.  I'd just maybe think about making it so that we can still have the distributed keyword without things breaking.  

Maybe we need to add in a better system for deprecating things?

rsciio/quantumdetector/_api.py
ericpre: Done!

PR #415, opened 2025-06-25
Add distributed lazy reading of npy file

ericpre:
While reading a `npy` file using `np.load` is trivial, this plugins implement distributed lazy reading of npy file.

### Progress of the PR
- [x] Implement distributed lazy reading of npy file,
- [x] add tests,
- [x] ready for review.


francisco-dlp (changes requested):
Code-wise, LGTM except for the minor coverage issue.

Since this file format does not support metadata, wouldn't it be good to add a warning when saving on this format, to make the users aware of this limitation?

ericpre:
> Since this file format does not support metadata, wouldn't it be good to add a warning when saving on this format, to make the users aware of this limitation?

In the interest of consistency between plugins, it will need to be done for all plugins, otherwise, it could be assumed that the absence of warning implies that the metadata are "supported". And best to keep this for a separate PR!

CSSFrancis (approved):
This looks pretty simple and good to me.  Loading data from `numpy` files is always helpful although I maybe wouldn't recommend saving in a `numpy` format :)

rsciio/numpy/_api.py
```diff
@@ -0,0 +1,185 @@
...
+    ----------
+    %s
+    %s
+    **kwargs : dict, optional
+        Additional keyword arguments passed to :func:`numpy.save`.
+    """
+    array = signal["data"]
+    if isinstance(array, da.Array):
+        raise TypeError("Lazy signal are not supported for writing to npy files.")
+
```
francisco-dlp, line 105: Won't you consider adding a warning here to remind the user that all metadata is lost when saving in this format for the users who don't read the User Guide?

ericpre: I added a note in the docstring. I don't think that we should warn here, because this is expected behaviour: the numpy  doesn't support metadata.
As mentioned above, if we want to go into this consideration, this should be done in a separate PR and done consistently with other plugins.

francisco-dlp: The reason why I think a warning is better is because it is more likely that an user who doesn't know about this limitation reads the warning than the docstring/user guide. However, I may be taking an overcautious appraoch to this. Therefore, let's leave it as is. In the unlikely event that this turns out to be a problem,  somebody will write an issue on it and we'll add the warning then. I'll take the same approach for my other comment.

francisco-dlp (approved):

rsciio/numpy/_api.py
ericpre: I have open https://github.com/hyperspy/rosettasciio/issues/421 to discuss it.

Issue #457, opened 2025-11-24
Security risk when loading PRZ files caused by using Pickle


uellue:
The PRZ file format for CEOS Panta Rhei uses pickled Python data structures to store metadata. RosettaSciIO uses NumPy with `allow_pickle=True`
https://github.com/hyperspy/rosettasciio/blob/4d5db8a8c7b7e3909b6d7eb9951dfbcd1fdb47fe/rsciio/pantarhei/_api.py#L59 to load such PRZ files.

Per the documentation of [`numpy.load()`](https://numpy.org/doc/stable/reference/generated/numpy.load.html) and [`pickle`](https://docs.python.org/3/library/pickle.html), that is not a good idea for untrusted data. An average user will not expect that loading a data file with `hs.load()` exposes them to arbitrary code execution.

For that reason unpickling should be deactivated unless explicitly switched on by the user, for example with sth like `hs.load(..., allow_pickle=True)`. If it is of interest, I could work on a PR to implement this.

@sk1p is already in contact with CEOS to work on changing Panta Rhei's data format so that it doesn't need unpickling.

Alternatively one could also look into using a safe parser for simple pickled data structures. In a quick search I didn't find much. Using JSON instead of pickled Python sounds like a better design to me.

uellue:
Ah I see, that only works in Gitlab! On Github, maintainers should be able to create a draft security advisory, probably here: https://github.com/hyperspy/rosettasciio/security/advisories If that works, one could delete the issue here.

CEOS just sent us a message yesterday that they would submit a patch to RosettaSciIO with a safer reader.

PR #459, opened 2025-12-04
Patch Panta Rhei prz file loading

tensionhead:
### Patch prz file loading
- addresses #457
- support Panta Rhei >=0.24.3 prz format where meta data will be loaded from a json string
- added `restricted_unpickling.py` module to prevent arbitrary code execution exploits via pickle when loading legacy prz files
- added `allow_restricted_pickle` flag:
  - ignored for prz files saved with Panta Rhei >=0.24.3
  - defaults to `False`: no unpickling, meta data can not be loaded for prz stored with Panta Rhei <0.24.3
  - if set to `True`: uses restricted pickle to load the meta data from legacy prz files
  

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs
# load prz without pickle 
# for prz stored with Panta Rhei >=0.24.3 the meta data gets loaded from a json string
# for prz stored with Panta rhei <0.24.3 loads without meta data
hs.load("panta_rhei_sample.prz")
# load legacy prz with using restricted pickle
hs.load("panta_rhei_sample_legacy.prz", allow_restricted_pickle=True)
```


rsciio/pantarhei/restricted_unpickling.py
```diff
@@ -0,0 +1,99 @@
...
+from packaging.version import Version
+
+numpy_version = Version(numpy.__version__)
+
+if numpy_version < Version("2.3"):
+    # for numpy < 2.3
+    from numpy.lib.format import _check_version, _read_array_header, read_magic
+else:
+    # for numpy >= 2.3
+    from numpy.lib._format_impl import _check_version, _read_array_header, read_magic
```
ericpre: Can we avoid importing private numpy API? There is some functionality that may be relevant in https://github.com/hyperspy/rosettasciio/blob/4d5db8a8c7b7e3909b6d7eb9951dfbcd1fdb47fe/rsciio/numpy/_api.py#L36C1-L37C1

tensionhead: I think using the existing code you pointed to at https://github.com/hyperspy/rosettasciio/blob/4d5db8a8c7b7e3909b6d7eb9951dfbcd1fdb47fe/rsciio/numpy/_api.py#L36C1-L37C1 is the best idea to avoid private numpy API calls. I see two options:
1. Slight refactoring of `inspect_npy_file` to also accept byte streams (i.e from an npz) and import into `pantarhei._api` module. Probably best to move it to `utils` to not make the pantarhei api dependent on the numpy api implementation.
2. Copying of the relevant parts to a new function like `inspect_npy_bytes` residing directly in the`restricted_unpickling`  module.

What would you suggest?

tensionhead:
@ericpre could you rerun the code coverage analysis? d223dd7 should have improved coverage.

ericpre (approved):
Thank you @tensionhead for the PR.

I fix a docstring that was failing the docstring test and privatise the `restricted_unpickling` module since it doesn't need to be public.

PR #464, opened 2025-12-09
App5 loader v3

argerlt:
Updated TopSpin loader PR without the formatting mistakes and repeatedly altered large binary test files.

Key Features:
        - App5 Reader (no writer, as it doesn't make sense here)
        - `dryrun` variable to quickly summarize large (10-100GB) app5 files without needing to load them
        - loads both 4D SPED and 2D  STEM taken with topspin
        - Because TopSpin Metadata is unique, harvests almost all info for 'axis' and 'metadata' directly from datasets.
        - requires h5py.
 
Also contains several of the missing features pointed out by @ericpre  and CoPilot in the previous PR #454. Unresolved comments from that PR are also noted below.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```
[In] >>> ]file_reader(file_from_export, dryrun=True)

[Out] >>> 
The following would be imported from /home/arg6/GitHub/rosettasciio/rsciio/tests/data/topspin/topspin_from_export.app5:
   - Test data B for rsciio, [3, 7, 29, 29]
      18d9446f-22bf-4fb1-8d13-338174e75d20/3526f008-a687-41fb-a21e-c21362241492
   - Test data B for rsciio, [11, 13]
      18d9446f-22bf-4fb1-8d13-338174e75d20/b0212ada-90b1-4117-8ed5-355717e4910b
   - Test data B for rsciio, [13, 17]
      18d9446f-22bf-4fb1-8d13-338174e75d20/b0e00442-4f83-4585-b84c-b82d2537b14b
   - Test data C for rsciio, [11, 13]
      b38446c6-6b25-4fee-9c62-6d8e4fcbeb5c/5a203dd8-ddbd-4775-8786-78d456e8b877```

### DIFF
--- a/rsciio/topspin/_api.py
+++ b/rsciio/topspin/_api.py
@@ -18,6 +18,7 @@
 
 
 import datetime
+import logging
 import mmap
 import os
 import sys
@@ -25,7 +26,6 @@ import xml.etree.cElementTree as ET
 
 import h5py
 import numpy as np
-from tqdm import tqdm
 
 from rsciio._docstrings import (
     FILENAME_DOC,
@@ -33,6 +33,9 @@ from rsciio._docstrings import (
     RETURNS_DOC,
     SHOW_PROGRESSBAR_DOC,
 )
+from rsciio.utils.tools import dummy_context_manager
+
+_logger = logging.getLogger(__name__)
 
 
 def file_reader(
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: rsciio/tests/data/topspin/create_topspin_data.py [hunk 1]
```diff
@@ -0,0 +1,235 @@
+# -*- coding: utf-8 -*-
```
argerlt: @ericpre These test files are now downsized versions of the originals, not brand new files. I agree, much better and easier to follow.

Let me know if you are happy with these. I have not been able to collect un-downsized files still, our setup is currently not available to us unfortunately. It might be a few months before I can get intentionally tiny datasets that don't require downsizing.

rsciio/tests/test_topspin.py
```diff
@@ -0,0 +1,169 @@
...
+        dims_str = [x.split("]")[0] for x in txt.split("[")[1:]]
+        assert dims_str == correct_sizes[i]
+
+
+def test_with_hyperspy():
+    hs = pytest.importorskip("hyperspy.api", reason="hyperspy not installed")
+    for f in [file_A, file_B, file_C]:
+        out = file_reader(f, show_progressbar=False)
+        for dset in out:
+            signal = hs.signals.Signal2D(
```
argerlt: https://github.com/hyperspy/rosettasciio/pull/454/files#r2552822555:
> Use hs.load as an additional check that the dictionary returned by file_reader is compliant.

rsciio/topspin/_api.py
```diff
@@ -0,0 +1,432 @@
...
+    Read .app5 file format used by NanoMegas's Topspin software.
+
+    .app5 files use the hdf5 file format, with metadata stored as
+    binarized XML-style text strings.
+
+
+    Parameters
+    ----------
+    %s
+    subset: str or None
```
argerlt: > I am not sure what is meant with "h5py-style address". Maybe some an example inline here or in the example section?

 "h5py Group name" is a better name perhaps? Its the location inside the hdf5 file where the data is stored. I will clarify.

rsciio/topspin/_api.py
```diff
@@ -0,0 +1,432 @@
...
+        address = task_list[i][0]
+        if len(shape) == 2:
+            # this progress bar is really just a "1-of-1", but it still
+            # reports load time and confirms loading succeeded
+            for dummy_iter in tqdm(
+                [1],
+                desc="Loading {}: ".format(address),
+                disable=not show_progressbar,
+            ):
+                datasets_list[i]["data"] = np.asanyarray(h5_file[address])
```
argerlt: > For the progress bar, I would suggest to use an outer progress bar with the increment changing when moving to the next dataset and an inner progress bar for loading the 4D dataset, then the "1 of 1 " progress bar can be removed.

> For both progress bar to play together, it may be necessary to use the position and clear keywords - see for example https://calmcode.io/course/tqdm/nested-loops.

Good suggestion, will do.

rsciio/topspin/_api.py
```diff
@@ -0,0 +1,432 @@
...
+
+        elif len(shape) == 4:
```
argerlt: https://github.com/hyperspy/rosettasciio/pull/454/files#r2552811344
@ericpre : 
> Do you have a file with a dataset with 3 dimension? Would it be possible to get one? If so, it can be done in a separate PR.



I don't think so.

Internally, the SPED datasets are stored like a vector of 2D diffraction patterns, (so, pseudo-3d), then reshaped into 4D during loading. Thus, I think even a 1x400x256x256 map would show up as a 4D signal.

rsciio/topspin/_api.py
```diff
@@ -0,0 +1,432 @@
...
+                data[int(key)] = h5_file[address][key]["Data"]
+            datasets_list[i]["data"] = data.reshape(shape)
+
+    return datasets_list
+
+
+file_reader.__doc__ %= (FILENAME_DOC, RETURNS_DOC)
+
+
+def _get_4D_axes(h5_grp):
```
argerlt: > The axes will also need the optional navigate entry to specify that this is a navigation dimension...

https://github.com/hyperspy/rosettasciio/pull/454/files#r2543183464

rsciio/tests/data/topspin/create_topspin_data.py
ericpre: Okay, if this is not possible to generate these dataset, then it is fine to keep the existing files.

rsciio/topspin/_api.py
argerlt: looks like `dataset_path` is the equivalent for the NX reader. Not an exact one-to-one, but close enough to justify mimicking the variable name. 

Fixed.

rsciio/topspin/_api.py
```diff
@@ -0,0 +1,432 @@
...
+import h5py
+import numpy as np
+from tqdm import tqdm
+
+from rsciio._docstrings import FILENAME_DOC, RETURNS_DOC
+
+
+def file_reader(filename, subset=None, dryrun=False, show_progressbar=True):
+    """
+    Read .app5 file format used by NanoMegas's Topspin software.
```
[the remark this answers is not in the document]
argerlt: Fixed.

rsciio/topspin/_api.py
argerlt: Fixed.

rsciio/topspin/_api.py
```diff
@@ -0,0 +1,432 @@
...
+            SPED_dtype = h5_file[address][first_key]['Data'].dtype
+            data = np.zeros([np.prod(shape[:-2]), shape[-2], shape[-1]], dtype=SPED_dtype)
+            key_count = len(h5_file[address].keys())
+            for key in tqdm(
+                h5_file[address],
+                desc="Loading {}: ".format(address),
+                total=key_count,
+                disable=not show_progressbar,
+            ):
+                data[int(key)] = h5_file[address][key]["Data"]
```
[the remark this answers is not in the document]
argerlt: Turns out the data isn't actually contiguous because of how the metadata is stored, but you still see a considerable (maybe 5x?) speedup by loading from the offsets using mmap and np.loadbuffer

It's fast enough I also removed the nested loading bar, which I think looks cleaner.

rsciio/topspin/_api.py
argerlt: fixed via removal of the nested loading bar (the loading is now much faster and I think it looks cleaner this way)

rsciio/topspin/_api.py
argerlt: fixed.

rsciio/tests/test_topspin.py
ericpre: There is still no test that uses `hs.load`, instead the hyperspy signals is constructed manually, which is different. Please add one.

The reason for checking against `hs.load` is that it would check that the signals axes and navigation are set correctly. This will also checked that the `signal_type` is set correctly because it is used to automatically return the correct pyxem signals, when pyxem is installed.

rsciio/topspin/_api.py
```diff
@@ -0,0 +1,482 @@
...
+import xml.etree.cElementTree as ET
+
+import h5py
+import numpy as np
+from tqdm import tqdm
+
+from rsciio._docstrings import FILENAME_DOC, RETURNS_DOC
+
+
+def file_reader(filename, subset=None, dryrun=False, show_progressbar=True):
```
ericpre: Mismatch between `subset` and `dataset_path`.

rsciio/topspin/_api.py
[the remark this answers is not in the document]
argerlt: I did try several variations of this for longer than I care to admit, but the metadata strings are always slightly different lengths:

```
IN    >>>   np.unique(np.array(offsets[1:])-np.array(offsets[:-1]))-(256*256*4)
OUT >>>  array([  1429,   3477,   4096,   4184,   4224,   4608,   4984,   6144,
         7800,  15360,  24576,  47104,  92824, 182816, 363160])
```

If there was a shortcut to predict these string lengths without reading them, I was not clever enough to find it

argerlt:
pre-commit.ci autofix

rsciio/tests/data/topspin/create_topspin_data.py
```diff
@@ -0,0 +1,236 @@
...
+def downsample_4D_group(group, mask):
+    """In-place reduction of hdf5 group representing 4D-STEM data."""
+    pxls = mask.flatten() * np.arange(mask.size)
+    pxls_dict = dict([str(x), str(i)] for i, x in enumerate(pxls[pxls > 0]))
+    gnames = [str(x) for x in np.arange(len(group.keys()))]
+    for gname in gnames:
+        if gname in pxls_dict.keys():
+            new_id = pxls_dict[gname]
+            group.move(gname, new_id)
+            data = group[new_id]["Data"][::16, ::16].copy()
```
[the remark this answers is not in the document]
ericpre: This is a common to have to figure what is the correct order of the array and axes. This can be checked easily by comparing the data displayed in the reference software and HyperSpy. Once this is checked, using an aspect-ratio different from 1:1 and checking the dimension of the array and axes is useful to avoid regression without having to check manually on the reference software, etc.

PR #488, opened 2026-03-06
Support for PRZ files with arbitrary dispersion direction

gguzzina:
Historically Panta Rhei used to store EELS data with an "energy type" dispersion, similar to what is usually done in XPS. Since Panta Rhei 0.23 this behavior is toggleable, and the overwhelming majority of users have switched to "energy loss type" behaviour, coherently with what is normally done in the EELS community.

This is not just a cosmetic change, the energy channels are actually in a different order in the stored files, and this requires a change in the loader.

### Progress of the PR
- [ ] Check formatting of the changelog entry (and eventual user guide changes) in the `docs/readthedocs.org:rosettasciio` build of this PR
- [ ] add tests,
- [ ] ready for review.

### DIFF
--- a/rsciio/pantarhei/_api.py
+++ b/rsciio/pantarhei/_api.py
@@ -144,8 +144,13 @@ def import_pr(data, meta_data, filename=None):
     content_type = meta_data.get("content.types")
     calibrations = []
     for axis in range(data_dimensions):
+        calib_key = "device.calib"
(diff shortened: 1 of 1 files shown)


ericpre (commented):
Thank you @gguzzina, after adding comments to the code, tests and a changelog entry, this will be all good to merge!

PR #491, opened 2026-03-10
Add Tofwerk TofDAQ HDF5 reader plugin for fibTOF FIB-SIMS

jat255:
## Summary

Adds a new read-only plugin for Tofwerk TofDAQ HDF5 files (`.h5`) from fibTOF FIB-SIMS instruments (Tescan plasma-FIB + Tofwerk ToF-SIMS detector). Closes #490.

### What's included

- **`rsciio/tofwerk/`** — new plugin with `__init__.py`, `_api.py`, `specifications.yaml`
- **Two reading modes** based on file type:
  - *Opened files* (PeakData present): returns 4D peak-integrated signal `(depth, y, x, m/z)` with calibrated mass axis
  - *Raw files* (FullSpectra/EventList): returns a sum spectrum and a TIC map, and optionally can compute a 4D peak-integrated signal
- **Tests** (`rsciio/tests/test_tofwerk.py`, 341 lines) with two synthetic test HDF5 files
- **Docs** (`doc/supported_formats/tofwerk.rst`) and index entry
- **Changelog** entry in `upcoming_changes/490.new.rst`


ericpre (commented):
From a quick look, this looks good.
If the two tests file have been generated by the python script (which seems to be the case), as opposed to created by the TofDAQ software, these files don't need to be added to the test data registry. They can simply be generated before running the test file and deleted after, as it is done for testing the empad reader: https://github.com/hyperspy/rosettasciio/blob/091d87318d1f1fda44d30a7e917797f8bd4f6334/rsciio/tests/test_empad.py.

I am surprised that the test in https://github.com/hyperspy/rosettasciio/blob/091d87318d1f1fda44d30a7e917797f8bd4f6334/rsciio/tests/test_import.py#L115 doesn't fail without editing it in the build with minimum dependencies on GitHub CI, because the same special case should apply as all reader that depends on h5py.

The table in doc/supported_formats/index.rst needs to be updated too.

rsciio/tofwerk/_api.py
```diff
@@ -0,0 +1,696 @@
...
+                    continue
+                event_masses = mass_axis[processed[valid]]
+                bin_idx = np.searchsorted(sorted_low, event_masses, side="right") - 1
+                in_peak = (
+                    (bin_idx >= 0)
+                    & (bin_idx < npeaks)
+                    & (event_masses <= sorted_high[bin_idx])
+                )
+                counts = np.bincount(bin_idx[in_peak], minlength=npeaks)
+                result[w, s, x, sort_idx] = counts
```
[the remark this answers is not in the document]
jat255: Yes, good idea. I changed this to use numba if available, and fall back to a partially-vectorized numpy implementation if numba is not installed. Both are faster than the first implementation

rsciio/tofwerk/_api.py
```diff
@@ -0,0 +1,836 @@
...
+            # Neither datetime source is available; return empty strings
+            _logger.warning(
+                "Could not parse acquisition time; date/time fields omitted."
+            )
+    return date_str, time_str, tz_str
+
+
+def _parse_pixel_size(file, nx):
+    """
+    Return pixel size in µm.
```
ericpre: Should it return the unit without conversion? Off the top of my head, units are not converted by default but this is not explicitly defined.

What do you think?

rsciio/tofwerk/_api.py
```diff
@@ -0,0 +1,836 @@
...
+
+    Source: ``FIBParams.ViewField`` (in m, converted to µm) divided by *nx*.
+
+    Note: ``ViewField`` in TofDAQ reflects the FIB deflector range set by the
+    operator at the time of acquisition.  It may represent the full-range
+    maximum rather than the actual scan area, in which case the returned value
+    will be an overestimate.  No higher-accuracy spatial calibration is stored
+    in the raw HDF5 file.
```
ericpre: I find this comment unclear. What is the "FIB detector"?
Does it mean that this is the field of view of a "survey electron image" acquired before acquiring the SIMS data and that the field of view of the SIMS data can be a smaller area from the survey image?

rsciio/tofwerk/_api.py
```diff
@@ -0,0 +1,836 @@
...
+        return value.tolist()
+    return value
+
+
+# ---------------------------------------------------------------------------
+# Public API
+# ---------------------------------------------------------------------------
+
+
+def file_reader(filename, lazy=False, compute_peak_data=False):
```
ericpre: It would be good to add the `chunks` argument  and pass it to the right place - see the reference docstring: https://github.com/hyperspy/rosettasciio/blob/091d87318d1f1fda44d30a7e917797f8bd4f6334/rsciio/_docstrings.py#L56-L59

Being able to specify the chunks size can be very useful.

rsciio/tofwerk/_api.py
```diff
@@ -0,0 +1,836 @@
...
+        ``PeakData/PeakTable`` dataset, reconstruct the 4-D peak-integrated
+        array from ``FullSpectra/EventList`` by applying the mass calibration
+        and integrating within each peak's defined window.  The result is
+        identical in structure to the ``PeakData/PeakData`` dataset written by
+        the Tofwerk proprietary software.  This option has no effect on opened
+        files (which already contain ``PeakData/PeakData``).  Default is False.
+
+    Returns
+    -------
+    list of dict
```
ericpre: Same as above, try to reuse docstring:
https://github.com/hyperspy/rosettasciio/blob/091d87318d1f1fda44d30a7e917797f8bd4f6334/rsciio/_docstrings.py#L117-L132

rsciio/tofwerk/_api.py
```diff
@@ -0,0 +1,836 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Public API
+# ---------------------------------------------------------------------------
+
+
+def file_reader(filename, lazy=False, compute_peak_data=False):
+    """
+    Read a Tofwerk TofDAQ HDF5 file.
```
ericpre: Maybe add a sentence mentioned that the "raw" and "opened" files.

rsciio/tofwerk/_api.py
```diff
@@ -0,0 +1,836 @@
...
+        # numba path: flatten each (write, segment) row into a contiguous int64
+        # buffer and dispatch to the JIT-compiled _accumulate_peak_counts kernel.
+        # result_sx is reused across rows to avoid repeated allocation.
+        # Avoids ~6 intermediate per-pixel arrays; benchmarked as ~19× faster than a
+        # per-pixel HDF5 read loop on a 2.2GB dataset.
+        result_sx = np.zeros((nx, npeaks), dtype=np.float32)
+        for w in range(nwrites):
+            for s in range(nsegs):
+                flat, offsets = _flatten_event_row(el[w, s, :])
+                result_sx[:] = 0
```
ericpre: This should already be zeros?
```suggestion
```

ericpre (commented):
Thank you @jat255, I have some more comments. 😃 

In the context of "raw" and "opened" files, the use of "opened" sounds inappropriate at first. Could we find a better word to describe what it is? "Processed" can be misleading too, but maybe something along the line of "pre-processed"?

There are a few comments around the use of "peak data" that may need some thoughts/discussions.

Reading (depth, x, y, m/z) data can use quite large amount of memory and I think that it would be good to do the following:
- add lazy support when `compute_peak_data=True`.
- add parameters to reduce the size of the data, as it is done for example in the [JEOL asw](https://hyperspy.org/rosettasciio/supported_formats/jeol.html) or [Velox emd](https://hyperspy.org/rosettasciio/supported_formats/emd.html#rsciio.emd.file_reader) files. It could by specifying the dtype, or selecting a range of m/z or depth.

Do you have access to the Totwerk analysis software? I am wondering if it would be worth adding basic support for writing, since there is already code doing that in the test suite. But this is for another day! 😅

jat255:
> Do you have access to the Totwerk analysis software? I am wondering if it would be worth adding basic support for writing, since there is already code doing that in the test suite. But this is for another day! 😅

I unfortunately do not have access to the software, so I've been doing this a bit blind, which is partially why the implementation has been a little all over the place as I've figured out the best way to arrange things!

rsciio/tofwerk/_api.py
```diff
@@ -0,0 +1,1541 @@
...
+    inv_sort_idx = np.argsort(sort_idx)
+
+    result = np.zeros((nwrites, nsegs, nx, npeaks), dtype=np.float32)
+
+    _logger.warning(
+        f"Reconstructing peak data from EventList "
+        f"({nwrites} depth slices × {nsegs} × {nx} pixels × {npeaks} peaks). "
+        f"This may take considerable time depending on the size of the file. "
+        f"Consider opening the file with Tofwerk software first for instant lazy loading."
+    )
```
ericpre: I don't think that this should be a warning and the last sentence is a bit odd, not usre if this is particularly useful. For users who have access to the software, they will know most likely know about this!

ericpre (commented):
I agree, this is in better shape and the documentation is clear!

Please see my comments.

ericpre (commented):
There are two comments left that were previously unanswered in the `if _use_numba` block - copying the link to the comment doesn't work.

I made some more comments, and I suspect that I made some more in the next round of review: it is a large PR and it is difficult to review in one go in a timely fashion.

rsciio/tofwerk/_api.py
```diff
@@ -0,0 +1,1261 @@
...
+                    chunks=chunks if chunks is not None else "auto",
+                )[depth_start:depth_stop]
+                if not already_sorted:
+                    peak_data = peak_data[:, :, :, sort_idx]
+                if mz_sel is not None:
+                    peak_data = peak_data[:, :, :, mz_sel]
+                if dtype is not None:
+                    peak_data = peak_data.astype(dtype)
+            else:
+                if already_sorted:
```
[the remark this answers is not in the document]
jat255: This isn't directly related to this comment, but for comparison (I can share the script if you want), I did some testing with different chunk shapes and prototypical operations, and the dask default was not really any worse than some of the options (maybe slightly slower in one type of operation than loading the whole spectrum in one chunk, but certainly acceptable):

```
File:    pfib_1301x128x128_AlTiPt.h5
Dataset: PeakData/PeakData

Shape:         (1301, 128, 128, 217)  (depth × y × x × m/z)
Dtype:         float32
Size on disk:  18.50 GB
Native chunks: (1, 128, 128, 1)  (1 depth slice × 1 m/z channel)

[1/5] ───────────────────────────────────────────────────────
Strategy: auto (dask heuristic, current default)
  Chunks:  shape/chunk=(45, 128, 128, 45)  n=145  132.7 MB/chunk
  A. create dask array:          0.32 ms
  B. arr[0].compute():           19.8 ms
  C. arr[:10].compute():         199.5 ms
  D. pixel depth profile:        9.12 s
  E. TIC (first 100 slices):     1.64 s  (1 rep)

[2/5] ───────────────────────────────────────────────────────
Strategy: native HDF5  (1, 128, 128, 1)
  Chunks:  shape/chunk=(1, 128, 128, 1)  n=282317  0.1 MB/chunk
  A. create dask array:          0.25 ms
  B. arr[0].compute():           73.1 ms
  C. arr[:10].compute():         387.0 ms
  D. pixel depth profile:        44.70 s
  E. TIC (first 100 slices):     3.80 s  (1 rep)

[3/5] ───────────────────────────────────────────────────────
Strategy: depth_slice  (1, 128, 128, 217)  [pre-146521b8 default]
  Chunks:  shape/chunk=(1, 128, 128, 217)  n=1301  14.2 MB/chunk
  A. create dask array:          0.24 ms
  B. arr[0].compute():           18.0 ms
  C. arr[:10].compute():         179.3 ms
  D. pixel depth profile:        9.27 s
  E. TIC (first 100 slices):     1.65 s  (1 rep)

[4/5] ───────────────────────────────────────────────────────
Strategy: ten_slices   (10, 128, 128, 217)
  Chunks:  shape/chunk=(10, 128, 128, 217)  n=131  142.2 MB/chunk
  A. create dask array:          0.14 ms
  B. arr[0].compute():           17.8 ms
  C. arr[:10].compute():         178.8 ms
  D. pixel depth profile:        9.40 s
  E. TIC (first 100 slices):     1.66 s  (1 rep)

[5/5] ───────────────────────────────────────────────────────
Strategy: row_slice    (1, 1, 128, 217)
  Chunks:  shape/chunk=(1, 1, 128, 217)  n=166528  0.1 MB/chunk
  A. create dask array:          0.29 ms
  B. arr[0].compute():           962.9 ms
  C. arr[:10].compute():         9.39 s
  D. pixel depth profile:        9.49 s
  E. TIC (first 100 slices):     93.12 s  (1 rep)

======================================================================================================

Strategy                                                    create   arr[0]    [:10]    pixel  TIC×100
------------------------------------------------------------------------------------------------------
auto (dask heuristic, current default)                     0.32 ms  19.8 ms 199.5 ms   9.12 s   1.64 s
native HDF5  (1, 128, 128, 1)                              0.25 ms  73.1 ms 387.0 ms  44.70 s   3.80 s
depth_slice  (1, 128, 128, 217)  [pre-146521b8 default]    0.24 ms  18.0 ms 179.3 ms   9.27 s   1.65 s
ten_slices   (10, 128, 128, 217)                           0.14 ms  17.8 ms 178.8 ms   9.40 s   1.66 s
row_slice    (1, 1, 128, 217)                              0.29 ms 962.9 ms   9.39 s   9.49 s  93.12 s
======================================================================================================
```

I think this justifies using the default "auto" chunking from dask.

ericpre (commented):
This PR starts to be difficult to follow. I will suggest to finish the few comments, particularly the ones on using `read_direct` and reducing memory usage, because there are important in term of performance.

jat255:
> This makes me wonder: should it be optional to return these additional peaks? Should it be a separate signals? This additional peaks is already in the dataset, if seems a bit silly to put it in the same signal! At the same if the user is only interested in specific peak and these were manually selected, maybe they want to see it separately?

I agree that these are all reasonable changes/additions, though I think due to the effort already put in, I would suggest at this point they be left as opportunities for future improvements, if users request them! Likewise for the lazy/non-lazy idea in your same comment.

PR #495, opened 2026-03-21
Updated the data shape and the way to parse scan size according to new EMPAD v1.2.2

matao1984:
Based on issues #493  
- The output data shape has been changed to [scan_y, scan_x, height, width]. 
- The axes order was updated accordingly.
- Updated the method to parse the scan size for new EMPAD versions v1.2.2 
        - In new EMPAD version, root.pix_x and pix_y no longer exist. The fov is stored under iom_measurements.full_scan_field_of_view.x and y. And they are always the same value, representing the larger dimension of the scan if the scan is non-square. Also added is a scale_factor under iom_measurements.full_scan_field_of_view.scale_factor. This factor is to match the EMPAD fov to the internal fov at the same magnification. However, the fov reported in XML is mistakenly recorded as the fov * scale_factor. Therefore, the true fov should be the reported value / scale_factor. Also, in new EMPAD versions, the EMPAD scan can be set to 0.2, 0.4, 0.6, 0.8 or 1 times of the internal fov. This is stored at scan_parameters.scan_size. Therefore, the correct scan pixel size should be fov / scale_factor * scan_size / max(scan_x, scan_y).
- Diffraction space calibration is no longer stored in the XML. So added an option to specify the q calibration in 1/nm. If None, the data will be in pixel scales. 
- Sometimes the EMPAD data contains NaN values that cause issues in downstream processing. Added an option to remove these NaN values by replacing them with zeros. Tested with non-square scan data from EMPAD v1.2.2.


rsciio/empad/_api.py
```diff
@@ -152,22 +158,42 @@ def file_reader(filename, lazy=False):
...
+                scales[i] = scale
+                units[i] = unit
+
         except Exception:
             _logger.warning("The scale of the navigation axes cannot be read.")
 
-    try:
-        pixel_size = float(om.root.iom_measurements.calibrated_pixelsize) * 1e9
+    # try:
+    #     pixel_size = float(om.root.iom_measurements.calibrated_pixelsize) * 1e9
```
ericpre: Since some files have this key, we should parse this information when available. Maybe use `q_calibration` first when not `None` and fallback to this when possible.

ericpre (commented):
Thank you @matao1984 for the pull request, the changes look good and the explanations and comments are useful.

To be merged, the following points will need to be addressed:
- we still need to support older version, can you please update accordingly, i.e. put back the code you commented with corresponding if block, etc.?
- it needs some tests to prevent future regression. As you will see in the [test file of the empad reader](https://github.com/hyperspy/rosettasciio/blob/30ca9dd5fe806f7c916555af2b0dd5bb6011a0fd/rsciio/tests/test_empad.py#L39-L47), we add the `xml` file and the `raw` is generated when running the test suite to avoid keeping large file in the repository.
- it will need a release notes, see [these instructions](https://github.com/hyperspy/rosettasciio/blob/main/upcoming_changes/README.rst) from the [PR templates](https://github.com/hyperspy/rosettasciio/blob/main/.github/PULL_REQUEST_TEMPLATE.md).

Some of it may appear tedious as a first contributor, but this is needed to be maintainable.

matao1984:
> > The test simply needs to be update, to reflect that the name of the axis are also swapped.
> > Are you planning to provide a small test file? If not, I could try to edit the xml file (for version 1.2.2) you provided to make a smaller dataset if you are happy that this xml file is added to the test suite? Avoiding editing the xml file would be better though! In any case, the raw file should be generated when running the test suite.
> 
> I was able to get on the scope and took a small test dataset. I am looking at the test codes. Some quick questions: When I checked the 1.2.0 file that was already in data folder, I don't see any difference than the file I took with 1.2.2, except for that the 1.2.0 is an square scan while mine is non-square. Does it make sense to remove the 1.2.0 file and just use the 1.2.2 file to reduce redundancy? Also, in test_read_map function it calls hyperspy.io.load and that fails the assertion with the axes order. Since this is for the rsciio, I think it makes more sense to just call the file_loader in rsciio.empad and assert the expected values in the returned dictionary. What is your opinion on this?

Actually I checked what it looks like in hyperspy/pyxem and was able to write the test codes with hyperspy.io.load. Ultimately rsciio will serve for hyperspy so maybe it is better to keep in this way?

ericpre:
> Actually I checked what it looks like in hyperspy/pyxem and was able to write the test codes with hyperspy.io.load. Ultimately rsciio will serve for hyperspy so maybe it is better to keep in this way?

This is fine to use hyperspy in the test suite. In principle, it shouldn't be the case but in practice a lot of the code from hyperspy and most of the tests haven't been updated accordingly since it was split out of hyperspy. The tests of reader that have been added since then don't use hyperspy - at least for most of them!

ericpre (approved):
Thank you @matao1984, this is all good. I tweaked a bit the docstring, release notes and fix some tests. The test data registry needed to be updated, in case you are interested in the details, there are some explanation in the contributing guide: https://hyperspy.org/rosettasciio/contributing.html#adding-and-updating-test-data

PR #506, opened 2026-04-20
Bug Fix: OverflowError: Python int too large to convert to C long on Windows

matao1984:
### Description of the change
The original code line 274 assigned the data with 
```
offset2 = header["Data_offset_2"]
```
On Windows system this may trigger the error:
```
Python int too large to convert to C long
```
Fixes:
```
offset2 = int(header["Data_offset_2"])
```
Root cause:

On Windows, CPython's implementation tries to convert the argument to a C long, which is a signed 32-bit integer even in 64-bit builds. Any value above 2^32 - 1 overflows a signed 32-bit C long, causing the above error. By converting the large value of the numpy scalar to python integer, CPython uses the correct 64-bit seek path that supports up to 2^64 -1 bytes.

### Progress of the PR
- [ ] add tests,
- [x] ready for review.

### DIFF
--- a/rsciio/blockfile/_api.py
+++ b/rsciio/blockfile/_api.py
@@ -242,6 +242,7 @@ def file_reader(filename, lazy=False, chunks="auto", endianness="<"):
             UserWarning,
         )
     header = sarray2dict(header)
+    # Convert the header value to python int to avoid potential overflow issues on 32-bit systems
     note = f.read(int(header["Data_offset_1"]) - f.tell())
     # It seems it uses "\x00" for padding, so we remove it
     try:
@@ -271,6 +272,7 @@ def file_reader(filename, lazy=False, chunks="auto", endianness="<"):
     # navigator = np.fromfile(f, dtype=endianness+"u1", shape=(NX, NY)).T
 
     # Then comes actual blockfile
+    # header["Data_offset_2"] is np.uint32. Converting to int to avoid potential overflow issues on 32-bit systems
     offset2 = int(header["Data_offset_2"])
     # Every frame is preceded by a 6 byte sequence
     # (AA 55, and then a 4 byte integer specifying frame number)
--- a/rsciio/blockfile/_api.py
+++ b/rsciio/blockfile/_api.py
@@ -271,7 +271,7 @@ def file_reader(filename, lazy=False, chunks="auto", endianness="<"):
     # navigator = np.fromfile(f, dtype=endianness+"u1", shape=(NX, NY)).T
 
     # Then comes actual blockfile
-    offset2 = header["Data_offset_2"]
+    offset2 = int(header["Data_offset_2"])
     # Every frame is preceded by a 6 byte sequence
     # (AA 55, and then a 4 byte integer specifying frame number)
     if not lazy:
--- a/rsciio/blockfile/_api.py
+++ b/rsciio/blockfile/_api.py
@@ -242,7 +242,7 @@ def file_reader(filename, lazy=False, chunks="auto", endianness="<"):
             UserWarning,
         )
     header = sarray2dict(header)
-    note = f.read(header["Data_offset_1"] - f.tell())
+    note = f.read(int(header["Data_offset_1"]) - f.tell())
     # It seems it uses "\x00" for padding, so we remove it
     try:
         header["Note"] = note.decode("latin1").strip("\x00")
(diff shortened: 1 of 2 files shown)


ericpre:
From reading the code, I understand that the value of this offset depends on the header and the size of the virtual image. It means that its value will not hit the overflow error? I don't understand why the offset will be larger than 2^32?
Can you share a file to reproduce the error?

matao1984:
> From reading the code, I understand that the value of this offset depends on the header and the size of the virtual image. It means that its value will not hit the overflow error? I don't understand why the offset will be larger than 2^32? Can you share a file to reproduce the error?

Yes. https://gigamove.rwth-aachen.de/download/b74266a05ced065170962341b468919d
The file is over 10GB. I only saw this happened on Windows with the reasons explained above.

ericpre:
Okay, this is a similar issue as https://github.com/hyperspy/rosettasciio/issues/480.

Here is an example to reproduce the issue when loading the data lazily:
```python
import hyperspy.api as hs
import numpy as np
import dask.array as da

fname = "test.blo"

shape = (500, 500, 144, 144)

data = da.ones(shape, dtype=np.uint8)
s = hs.signals.Signal2D(data)
s.save(fname, overwrite=True)

s2 = hs.load(fname, lazy=True)
s2.compute()
```
Which give the following error:
```python
---------------------------------------------------------------------------
OverflowError                             Traceback (most recent call last)
Cell In[3], line 1
----> 1 s2.compute()

File ~\Dev\hyperspy\hyperspy\_signals\lazy.py:244, in LazySignal.compute(self, close_file, show_progressbar, **kwargs)
    196 def compute(self, close_file=False, show_progressbar=None, **kwargs):
    197     """
    198     Attempt to store the full signal in memory.
    199 
   (...)    242 
    243     """
--> 244     self.data = dask_utils._compute(
    245         self.data, show_progressbar=show_progressbar, **kwargs
    246     )
    247     if close_file:
    248         self.close_file()

File ~\Dev\hyperspy\hyperspy\misc\_dask_utils.py:192, in _compute(arrays, store_to, show_progressbar, **kwargs)
    190     da.store(arrays, store_to, compute=True, lock=False, **kwargs)
    191 else:
--> 192     return da.compute(arrays, **kwargs)[0]

File ~\miniforge3\Lib\site-packages\dask\base.py:685, in compute(traverse, optimize_graph, scheduler, get, *args, **kwargs)
    682     expr = expr.optimize()
    683     keys = list(flatten(expr.__dask_keys__()))
--> 685     results = schedule(expr, keys, **kwargs)
    687 return repack(results)

File ~\Dev\rosettasciio\rsciio\utils\_distributed.py:165, in slice_memmap(slices, file, dtypes, shape, key, positions, **kwargs)
    131 """
    132 Slice a memory mapped file using a tuple of slices.
    133 
   (...)    162     Array of the data from the memory mapped file sliced using the provided slice.
    163 """
    164 slices_ = np.squeeze(slices)[()]
--> 165 data = np.memmap(file, dtypes, shape=shape, **kwargs)
    166 if key is not None:
    167     data = data[key]

File ~\miniforge3\Lib\site-packages\numpy\_core\memmap.py:284, in memmap.__new__(subtype, filename, dtype, mode, offset, shape, order)
    281     acc = mmap.ACCESS_WRITE
    283 start = offset - offset % mmap.ALLOCATIONGRANULARITY
--> 284 bytes -= start
    285 # bytes == 0 is problematic as in mmap length=0 maps the full file.
    286 # See PR gh-27723 for a more detailed explanation.
    287 if bytes == 0 and start > 0:

OverflowError: Python int too large to convert to C long
```
 I was surprised that `offset` will cause this overflow issue, because the value of `offset` is never large enough but it must be that for large files, `bytes` is large enough to trigger the overflow because of the subtraction with `offset` or something related.

ericpre (commented):
This looks good, can you please add a comment in the code to explain why it is needed and a release note that explains what does it fixes?

matao1984:
> This looks good, can you please add a comment in the code to explain why it is needed and a release note that explains what does it fixes?
 
I have done some in-depth debugging and found that even with my 10GB data the offset2 is 644096, smaller than 2^31-1, thus the lazy loading was OK. The error happened at da.compute(). It seems to be related to some intermediate type conversion during the Dark computation regarding the memmap operations. It is not essentially the offset2 is too large, but on big data it can get large enough that triggers some intermediate values go beyond the limit. I also spotted line 246 which involves similar computations mixed np.uint32 and python int so I did the same python int conversion here to avoid potential problems. The test code ran without a problem.

ericpre (approved):

PR #508, opened 2026-04-20
Update NCEM emd reader to handle py4DSTEM data saved with EMD v1.0

matao1984:
### Description of the change
- Modified the axes parsing logic to look for dim0, dim1, dim2, dim3 in py4DSTEM data
- Modified the scale parser to return python float value instead of numpy value
- Modified the units parser to read the correct units and correctly handle angstrom unit (A vs Å)
- Added a small py4dstem file with data shape of 10x20x110x100 (Ry, Rx, Qy, Qx)
- Modified the test codes for the added data file

### Progress of the PR
- [x] add tests,
- [x] ready for review.


rsciio/tests/test_emd_ncem.py
```diff
@@ -110,6 +110,24 @@ def test_data_axis_length_1():
     signal = hs.load(filename)
     assert signal.data.shape == (5, 1, 5)
 
+@pytest.mark.parametrize("lazy", (True, False))
+def test_py4dstem(lazy):
+    filename = TEST_DATA_PATH / "py4DSTEM_size20x30x110x100.h5"
```
ericpre, line 115: Use a smaller file and non-square dimension in signal space? Size of 2, 3, 4, 5 will be enough.

rsciio/blockfile/_api.py
```diff
@@ -271,7 +271,7 @@ def file_reader(filename, lazy=False, chunks="auto", endianness="<"):
     # navigator = np.fromfile(f, dtype=endianness+"u1", shape=(NX, NY)).T
 
     # Then comes actual blockfile
-    offset2 = header["Data_offset_2"]
+    offset2 = int(header["Data_offset_2"])
```
ericpre, line 274: To avoid conflict with #506, remove this commit?

rsciio/tests/test_emd_ncem.py
matao1984: Done

matao1984:
@ericpre The test file has been replaced and the test codes modified accordingly.

I noticed that for some py4DSTEM files saved together with "virtualimage", "dp_mean", etc, reading with default "stack_group=None" did not work. It needs to be set to "False" in order to read all the datasets correctly. I did not modify this logic but added some notes in the doc string in the file_reader function.

ericpre:
pre-commit.ci autofix

ericpre (approved):

PR #505, opened 2026-04-12
Increase tolerance in digitalsurf

ericpre:
Follow up of #492: increase tolerance further as noticed in https://github.com/hyperspy/hyperspy-extensions-list/actions/runs/24273390216/job/70882652568. The error depends on compression losses and it seems to be environment dependent, which makes difficult to control the error and therefore to figure out what is a good tolerance in the test!

### Progress of the PR
- [x] Increase tolerance in digitalsurf test,
- [n/a] update docstring (if appropriate),
- [n/a] update user guide (if appropriate),
- [n/a] add a changelog entry in the `upcoming_changes` folder (see [`upcoming_changes/README.rst`](https://github.com/hyperspy/rosettasciio/blob/main/upcoming_changes/README.rst)),
- [n/a] Check formatting of the changelog entry (and eventual user guide changes) in the `docs/readthedocs.org:rosettasciio` build of this PR (link in github checks)
- [x] update tests,
- [x] ready for review.

PR #510, opened 2026-04-25
Increase tolerance in digitalsurf

ericpre:
Similar as #505, more tolerance increase, as this is still occurring as for example in https://github.com/hyperspy/hyperspy-extensions-list/issues/98. Do it more this time!

### Progress of the PR
- [x] Increase tolerance in digitalsurf test,
- [n/a] add a changelog entry in the `upcoming_changes` folder (see [`upcoming_changes/README.rst`](https://github.com/hyperspy/rosettasciio/blob/main/upcoming_changes/README.rst)),
- [n/a] Check formatting of the changelog entry (and eventual user guide changes) in the `docs/readthedocs.org:rosettasciio` build of this PR (link in github checks)
- [x] update tests,
- [x] ready for review.

### DIFF
--- a/rsciio/tests/test_digitalsurf.py
+++ b/rsciio/tests/test_digitalsurf.py
@@ -597,8 +597,8 @@ def test_writetestobjects(tmp_path, test_object):
     d2.save(fn, is_special=False)
     d3 = hs.load(fn)
 
-    np.testing.assert_allclose(d2.data, d.data, rtol=1e-6)
-    np.testing.assert_allclose(d2.data, d3.data)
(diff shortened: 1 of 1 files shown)


ericpre:
I suggest to ignore the suggestion from co-pilot because I don't think that they are really needed here: the test files don't change, the issue is related to the use of compression that varies across platform and I think that increasing tolerance is sufficient here.

CSSFrancis (approved):

CSSFrancis:
Looks good. @ericpre I'll let you merge. With #518 merged the docs should be passing now.

PR #515, opened 2026-05-15
Improve DE MRC File Loading

CSSFrancis:
This pull request significantly refactors and improves the handling and parsing of DE movie and related files in the `rsciio.mrc` module. The main changes include the introduction of robust regex-based filename parsing, a new utility function to gather related files, and a reorganization of how virtual and external images are loaded and stored in metadata. These updates make the code more maintainable, extensible, and accurate in handling various file naming conventions.

**Improvements to DE file handling and metadata extraction:**

* Added regular expressions (`MOVIE_RE`, `VIRTUAL_RE`) to robustly parse DE movie and virtual image filenames, supporting multiple naming conventions.
* Introduced the `find_related_de_files` function to reliably discover and deduplicate related info files, virtual images, external images, and scan CSVs associated with a DE movie file.
* Refactored the `file_reader` function to use `find_related_de_files`, improving reliability and simplifying logic for finding metadata, virtual, and external images.

### DIFF
--- a/rsciio/mrc/_api.py
+++ b/rsciio/mrc/_api.py
@@ -380,10 +380,12 @@ def file_reader(
                     suffix = ""
                 if len(dir_name) == 0:
                     dir_name = "."
-                metadata = glob.glob(dir_name + "/" + unique_id + suffix + "_info.txt")
-                virtual_images = glob.glob(
-                    dir_name + "/" + unique_id + suffix + "*_[0-4]_*.mrc"
-                )
+                metadata = glob.glob(dir_name + "/" + unique_id + suffix + "*_info.txt")
+                virtual_images = [
+                    p
+                    for p in glob.glob(dir_name + "/" + unique_id + suffix + "*_[0-4]_*.mrc")
+                    if "movie" not in os.path.basename(p).lower()
+                ]
                 virtual_images_old = glob.glob(
                     dir_name + "/" + unique_id + suffix + "_virt[0-4]_*.mrc"
                 )
@@ -434,24 +436,27 @@ def file_reader(
         positions = None
     metadata["General"]["original_filename"] = os.path.split(filename)[1]
     metadata["Signal"]["signal_type"] = ""
-
+    metadata["General"]["virtual_images"] = {}
     if virtual_images is not None and len(virtual_images) > 0:
         imgs = []
-        for v in virtual_images:
-            img = file_reader(v)[0]["data"]
+        for i, v in enumerate(virtual_images):
+            signal = file_reader(v)[0]
+            img = signal["data"]
             if navigation_shape is not None and navigation_shape[::-1] != img.shape:
                 if np.prod(img.shape) == np.prod(navigation_shape[::-1]):
-                    img = img.reshape(navigation_shape[::-1])
+                    signal["data"] = img.reshape(navigation_shape[::-1])
                 else:  # pragma: no cover
                     _logger.warning(
                         f"Virtual image {v} does not match the navigation shape {navigation_shape[::-1]}"
                     )
-            imgs.append(img)
-        metadata["General"]["virtual_images"] = imgs
+            imgs.append(signal)
+            print(f"Adding virtual image _sig_image_{i}")
+            metadata["General"]["virtual_images"][f"_sig_image_{i}"] = signal
+            print(metadata["General"]["virtual_images"])
         # checking to make sure the navigator is valid
-        if navigation_shape is not None and navigation_shape[::-1] == imgs[0].shape:
+        if navigation_shape is not None and navigation_shape[::-1] == imgs[0]["data"].shape:
             metadata["_HyperSpy"] = {}
-            metadata["_HyperSpy"]["navigator"] = imgs[0]
+            metadata["_HyperSpy"]["_sig_navigator"] = imgs[0]
 
     if external_images is not None and len(external_images) > 0:
         imgs = []
--- a/rsciio/mrc/_api.py
+++ b/rsciio/mrc/_api.py
@@ -606,17 +606,17 @@ def file_reader(
 
     original_metadata["std_header"] = sarray2dict(std_header)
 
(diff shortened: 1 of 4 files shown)


CSSFrancis:
pre-commit.ci autofix

rsciio/mrc/_api.py
```diff
@@ -434,30 +530,48 @@ def file_reader(
...
                     )
-            imgs.append(img)
-        metadata["General"]["virtual_images"] = imgs
+            imgs.append(signal)
+            _logger.debug(f"Adding virtual image _sig_image_{i}")
+
+            if virtual_image_names and i < len(virtual_image_names):
+                name = f"_sig_{virtual_image_names[i]}"
+            else:
+                name = f"_sig_Virt_{i}"
```
ericpre: Add a comment here explaining the naming used?

rsciio/mrc/_api.py
```diff
@@ -434,30 +530,48 @@ def file_reader(
...
+            metadata["General"]["virtual_images"][name] = signal
         # checking to make sure the navigator is valid
-        if navigation_shape is not None and navigation_shape[::-1] == imgs[0].shape:
+        if (
+            navigation_shape is not None
+            and navigation_shape[::-1] == imgs[0]["data"].shape
+        ):
             metadata["_HyperSpy"] = {}
-            metadata["_HyperSpy"]["navigator"] = imgs[0]
+            metadata["_HyperSpy"]["_sig_navigator"] = imgs[0]
```
ericpre, line 562: This should not set the image as the navigator by default according to https://github.com/hyperspy/hyperspy/blob/6b303ec4be48a446233d75c5605de319d2cf8b2d/hyperspy/signal.py#L3009-L3010? However, the test suite still pass... Any idea why?

rsciio/mrc/_api.py
```diff
@@ -434,30 +530,48 @@ def file_reader(
...
         for e in external_images:
-            imgs.append(file_reader(e)[0]["data"])
-        metadata["General"]["external_detectors"] = imgs
+            imgs.append(file_reader(e)[0])
+        for i, signal in enumerate(imgs):
+            _logger.debug(f"Adding external image _sig_ext_{i}")
+            if external_images_names and i < len(external_images_names):
+                name = f"_sig_{external_images_names[i]}"
+            else:
+                name = f"_sig_ext_{i}"
```
ericpre: Same as above, add a comment to explain naming.

ericpre (commented):
Would it be possible to document the naming convention of external/virtual images.

The usage of `external_detectors` user guide needs to be updated.

CSSFrancis:
@ericpre yea I can document them. 

Probably should be something upstream in hyperspy as well. 
Virtual Images are calculated from some ROI and method.

External Images are from an external detector like a HAADF or BF detector.

I was thinking something like adding a dict of navigators. So you'd have something like `s.navigators["Virtual Bright Field"]`


Then maybe you could `s.plot(navigator="Virtual Bright Field")`

I guess we just need a formalization for the metadata format and then go from there.

rsciio/mrc/_api.py
CSSFrancis: @ericpre The `_sig_` is removed and the underlying dictionary is cast into a signal.  This was necessary because otherwise the navigator was a numpy array which caused problems.

CSSFrancis: https://github.com/hyperspy/hyperspy/blob/6b303ec4be48a446233d75c5605de319d2cf8b2d/hyperspy/misc/_utils.py#L469-L507

rsciio/mrc/_api.py
CSSFrancis: https://github.com/hyperspy/hyperspy/blob/6b303ec4be48a446233d75c5605de319d2cf8b2d/hyperspy/misc/_utils.py#L469-L507

CSSFrancis:
The failing doc links are fine but don't like the CI

rsciio/mrc/_api.py
ericpre: Okay, maybe worth adding as a comment for now until properly handled, as you mentioned for example in https://github.com/hyperspy/rosettasciio/pull/515#issuecomment-4483860119.

CSSFrancis:
@ericpre how does this format look:

Base Node: `signal.metadata._Hyperspy.navigators` 

Each navigator is then a seperate dictionary input i.e. (`signal.metadata._Hyperspy.navigators.Virtual_Bright_Field` which is set to a Signal2D with the same Dimensions as the Navigation Axes. 

These are acessible via `signal.navigators` which returns the dictiony of all of the different navigation signals.

ericpre:
Yes, it will works to set them as `navigators` attribute of signals, however, will it work when plotting? By default the `navigator` is not None, it will be picked up and it will expect a `BaseSignal`, which may not work if this is a dictionary?

rsciio/_load.py
ericpre, line 1: Is this file supposed to be part of the PR?

ericpre (approved):
This looks good, except the `_load.py` file which doesn't seem to be part of this PR?

rsciio/_load.py
CSSFrancis: No that should be for a different PR. Let me exclude that from the git history. I was going to make a generic `load` function like the hyperspy one that just returns the dictionaries.

CSSFrancis:
pre-commit.ci autofix

ericpre:
I will let you merge when you are happy with it - in case you want to tidy the git history, it seems that some commit repeat themselves.

CSSFrancis:
pre-commit.ci autofix

CSSFrancis:
@ericpre I added a test and a change because the json serialization wasn't handling some of the mrc bytes properly.  And I editing the commit history because it was pretty messy.  Feel free to merge if you'd like.

Also not sure why pre-commit is fighting me so much today.

CSSFrancis:
pre-commit.ci autofix

Issue #513, opened 2026-05-05
EMPAD reader parsing scan sizes


matao1984:
#### Describe the bug
The current EMPAD reader parses the scan sizes (for version >= v1.2.2) from these xml lines:
```
<scan_parameters mode="search">
<scan_resolution_y>64</scan_resolution_y>
<scan_resolution_x>64</scan_resolution_x>
<scan_size>1.0</scan_size>
<scan_center_y>0.5</scan_center_y>
<scan_center_x>0.5</scan_center_x>
</scan_parameters>
<scan_parameters mode="acquire">
<scan_resolution_y>256</scan_resolution_y>
<scan_resolution_x>256</scan_resolution_x>
<scan_size>0.6</scan_size>
<scan_center_y>0.5</scan_center_y>
<scan_center_x>0.5</scan_center_x>
</scan_parameters>
```
When xml.convert_xml_to_dict handle these keys it does not differentiate the "mode=" tag so later keys with mode="acquire" overwrites the mode="search" and gets returned as the scan size. This causes problem for data acquired using the "search" mode for wrong scan sizes. We will have to think a better way to parse the scan sizes. Maybe try to match the raw file name with this line?
```
<raw_file filename="scan_x256_y256.raw"/>
```

PR #520, opened 2026-05-27
Parsing scan sizes of EMPAD data

matao1984:
For issue #513 
### Description of the change
Modified the parsing logic for the scan sizes. Now the reader looks for the raw file name with a pattern of "scan_xaaa_ybbb.raw", in which the "aaa" and "bbb" are the scan sizes for x and y, respectively. All tests passed.

The reason for this change:

The EMPAD metadata (xml) saves two sets of scan parameters tagged "search" and "acquire" with the exactly same keys. When _parse_xml reads these keys it does not distinguish the different tags so the "search" keys are overwritten by the "acquire" keys. The original logic reads the scan sizes from these keys. Therefore, the parsed scan sizes are always from the "acquire" parameters. This may work most of the time because users usually use "acquire" to take the datasets. However, it will fail if the actual dataset has been taken with the "search" parameters (which the current EMPAD GUI does not forbid you to). On the other hand, at least up to the current version of EMPAD GUI v1.2.2 the raw file name is always in the pattern of "scan_xaaa_ybbb.raw". So this should be a safer way to parse the scan sizes.

### Progress of the PR
- [ ] add tests,
- [x] ready for review.

### DIFF
--- a/rsciio/empad/_api.py
+++ b/rsciio/empad/_api.py
@@ -80,16 +80,20 @@ def _parse_xml(filename):
     elif om.has_item("root.pix_x") and om.has_item("root.pix_y"):
         # 2D x 2D
         info.update({"scan_x": int(om.root.pix_x), "scan_y": int(om.root.pix_y)})
-    # in case root.pix_x and root.pix_y are not available
-    elif om.has_item("root.scan_parameters.scan_resolution_x") and om.has_item(
-        "root.scan_parameters.scan_resolution_y"
-    ):
-        info.update(
-            {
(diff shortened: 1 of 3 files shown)


ericpre (commented):
This looks good - I will double check with some old files but it seems that this is already covered in the test suite.

The tests are checking the shape of the array but the shape of the navigation and signal and I think that it would be good to check these too! Can you please update the tests accordingly?

matao1984:
Sorry for being late. I have added the test codes for both signal and navigation sizes. All test passed on my end.

ericpre (approved):

PR #525, opened 2026-06-23
feat(zspy): consolidate metadata for cloud-optimized I/O

francisco-dlp:
## Summary

After writing a zspy file, HyperSpy's hierarchical writer creates ~20+ separate zarr metadata files (\, \, \ per group, per axis, per metadata subgroup). On cloud storage (S3, GCS, Azure), each of these files requires a separate HTTP GET on open, adding significant latency before any data is accessed.

This PR adds \ support to the zspy writer, which bundles all metadata into a single \ file (zarr v2) or inlines into \ (zarr v3), reducing cloud open time from 20+ HTTP requests to 1.

### \
- New \ helper (lines 163-177) — handles both zarr v2 (\) and v3 (\)
- New \ parameter on \ (line 189) — enabled by default
- Called after \ (lines 299-307) in a try/except block — logs a warning on failure (store doesn't support writes) but the file remains valid

### \
- New \ class with 7 tests (lines 298-393):
  - Metadata creation/absence with consolidate True/False/default
  - Round-trip data + metadata integrity (both settings)
  - Complex nested metadata (axes, General/Signal metadata)
  - Cross-format comparison

### \
- Changelog entry for the new feature

## Context

This is part of a broader effort to add cloud storage I/O support to HyperSpy/rsciio. The chunk size (100 MB) and compression (Blosc zstd) in zspy are already cloud-ready — consolidated metadata was the remaining bottleneck.

## Note

The new tests use \ directly (a \ supported in both zarr v2 and v3). They cannot currently be collected due to pre-existing zarr v3 incompatibility in the test suite (\ was removed in zarr v3) — this is a separate issue, not caused by this PR.

### DIFF
--- a/rsciio/tests/test_zspy.py
+++ b/rsciio/tests/test_zspy.py
@@ -293,3 +293,101 @@ def test_save_store_error(tmp_path):
         ValueError, match="The `store_type` parameter must be None if a zarr "
     ):
         s.save(filename=store, store_type="zip")
+
+
+class TestConsolidateMetadata:
+    """Tests for consolidated metadata in ZSPY files."""
+
+    @pytest.fixture
+    def signal(self):
+        data = np.ones((10, 10, 10, 10))
+        s = hs.signals.Signal1D(data)
+        s.metadata.General.title = "test consolidate metadata"
+        return s
+
+    def test_consolidate_default_creates_zmetadata(self, signal, tmp_path):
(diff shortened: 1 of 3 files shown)


ericpre (commented):
This looks good. The documentation failure is genuine and the release note needs updating.

ericpre (commented):
Dummy comment to trigger the integration tests workflow

PR #138, opened 2023-06-12
Trivista xml2dict update

pietsjoh:
### Description of the change
The trivista filereader used its own version to convert an ET to a dict.
This PR switches this converter to utils.tools.XmlDict.

### Progress of the PR
- [x] Change implemented (can be split into several points),
- [x] add an changelog entry in the `upcoming_changes` folder (see [`upcoming_changes/README.rst`](https://github.com/hyperspy/rosettasciio/blob/main/upcoming_changes/README.rst)),
- [x] add tests,
- [ ] clarify how to read attrib only (see https://github.com/hyperspy/rosettasciio/issues/113#issuecomment-1587458669)
- [ ] ready for review.

PR #531, opened 2026-07-09
Trivista xml2dict update #2

janisschumacher:
### Description of the change
This PR finishes the unfinished PR #138 .

The Trivista Reader now uses the XmlDict class to dictionarize the XML Element Tree.
For that the parameter ignore_children was added to the  dictionarize method, so that its possible to convert just the attributes of a node and ignore the children.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs
signal = hs.load("SI-Trivista-Testspectrum.tvf")


```


ericpre (commented):
The comments of the co-pilot review are sensible and need addressing.

ericpre (approved):
Thank you @janisschumacher for the finishing this work. This is all good now. I rebased to get some CI workflow to pass and tidy up some code, for example the documentation build was failing because of warning related to docstring formatting.

There is a CI failure but this is not related to this PR and should be fixed in #533.

Issue #352, opened 2025-01-10
Tracking Zarr 3.0.0 Performance and Integration


CSSFrancis:
#### Describe the functionality you would like to see.

I'm going to try to implement support for zarr 3.0.0 over the next couple of days.  I'll try to test performance here and see if I can make a small guide to optimizing performance.  Specifically, I want to look at optimal sharding for 4D datasets for helping with efficient data slicing and to improve storage on windows computers.  I'm not sure how `zarr` + `dask` + sharding will ultimately preform but I assume that `dask` is not quite smart enough to handle that effectively, however, if the perfromance is good enough it might be worth returning the zarr array rather than automatically converting it to a dask array and only doing the conversion when necessary. 

As far as implementation goes:

- [ ] Implement [Local Store](https://zarr.readthedocs.io/en/latest/api/zarr/storage/index.html#zarr.storage.LocalStore)
- [ ] Implement [ZipStore](https://zarr.readthedocs.io/en/latest/api/zarr/storage/index.html#zarr.storage.ZipStore)
- [ ] Compare Speed for v2.0.0 to v3.0.0
    - [ ] Mac OS
    - [ ] Windows
- [ ] Test Sharding Implementation 
- [ ] Test GPU Direct Storage performance

ericpre:
Indeed, it is worth making the distinction between ragged array (array with variable length axis of "standard dtype") and object dtype.
Assuming that is possible to get zarr 3 to work with ragged array of standard dtype, other potential issues may be:
- Models and components, almost of them should be fine, because they are converted to python dictionary which are then saved to hdf5/zarr array or attributes. I am not sure about custom components (not implemented in hyperspy or extensions) which are not expression-based, because they may be pickled or something similar...
- array of string dtype?

CSSFrancis:
> Indeed, it is worth making the distinction between ragged array (array with variable length axis of "standard dtype") and object dtype. 

This seems realtively easy, there is already the [VLenBytesCodec](https://zarr.readthedocs.io/en/stable/_modules/zarr/codecs/vlen_utf8.html#VLenBytesCodec) which seems like it could be easily extended.  Although like the UTF encoding it isn't directly supported.

> Assuming that is possible to get zarr 3 to work with ragged array of standard dtype, other potential issues may be:
[...]

PR #536, opened 2026-07-24
Add CSR (dense buffer + offsets) encoding for uniform ragged arrays

CSSFrancis:
### Description of the change

Ragged arrays (`dtype=object`, one variable-length cell per navigation
position) are currently always written one independent object-dtype/vlen
blob per cell (`zarr`'s `VLenArray`/`MsgPack` object codec, or h5py's
`vlen_dtype`). That's necessary for genuinely heterogeneous data (ragged
strings, mixed dtypes, varying per-cell shapes), but it's overkill for the
very common case of *numeric ragged arrays where every cell shares the same
trailing shape and only the leading (count) dimension varies* — e.g. one
`(N_i, 2)` array of vector/point coordinates per position, which is exactly
what `Points`/`Lines`/`Rectangles` markers and diffraction-vector signals
store.

This PR adds an automatic, opt-in-by-shape alternative encoding for that
case:
- A single dense `values` array: every cell's rows concatenated along axis
  0, in the same flattened (row-major) nav order the existing
  `flatten_data` helper already walks. Since it's a concrete (non-object)
  dtype, it's chunked and compressed exactly like any ordinary dataset —
  no per-cell codec involved at all.
- A small `_ragged_offsets_<key>` sidecar: the cumulative per-cell row
  count (one `int64` per navigation position, not per vector), used to
  slice `values` back into cells on read.
- A per-dataset `_ragged_encoding="csr"` attribute so the reader knows
  which reconstruction path to use; datasets without it keep going through
  the existing generic path unchanged, so old files and non-conforming
  ragged data are unaffected.

The encoding is selected automatically from the actual data shape at write
time — no API change for callers (HyperSpy/pyxem don't need to opt in).
Detection and both write/read paths live entirely in
`rsciio/_hierarchical.py`, shared by both `hspy` (h5py) and `zspy` (zarr).

Because this is a new on-disk format, the hspy/zspy format version is
bumped to **3.4**. Readers predating it don't know about
`_ragged_encoding` and would otherwise silently misread a CSR-encoded
dataset as a plain dense array; the bump makes them emit the existing
"written using a newer version of the file format" warning.

Rewriting a key also clears any ragged bookkeeping from the previous
write (marker attributes and both sidecar flavours), so switching a key
between the two encodings — in either direction — is correct and doesn't
leave orphan sidecars behind.

**Known limitations / follow-ups**, flagged for discussion rather than
silently deferred:
- Lazy (dask-backed) ragged *writes* still go through the existing
  generic per-cell encoding; only eager writes take the new CSR path.
  Making the write itself lazy needs dask's "unknown chunk size" machinery
  and felt like a separate, larger piece of work.
- Read-side lazy reconstruction currently chunks along the outermost
  navigation axis only (each dask block = one or more whole "rows" of the
  nav grid). Correct for any navigation shape, just coarser-grained than
  optimal for very wide navigation grids with small physical `values`
  chunks.
Opening as a draft since this introduces a new (if narrowly-scoped and
backward-compatible) on-disk representation, and I'd like input on the
approach — in particular whether the encoding-selection heuristic and
attribute name are reasonable — before finishing off the follow-ups above.

---

### Measurements

Ragged diffraction-vector-shaped data: one `(N_i, 2)` float64 array per
navigation position, `N_i` uniform-random. Best of 3 runs, macOS, against
`upstream/main` at the same commit this branch is based on. Load is an
eager `hs.load()` of the whole ragged array.

| config | metric | main | this PR | |
|---|---|---|---|---|
| **zspy** 128×128, 318k vectors | size | 4749.9 KB | 4439.4 KB | −6.5% |
| | save | 57.2 ms | 27.4 ms | **2.09× faster** |
| | load | 47.1 ms | 19.2 ms | **2.46× faster** |
| **zspy** 64×64, 101k vectors | size | 1508.8 KB | 1412.8 KB | −6.4% |
| | save | 24.4 ms | 17.8 ms | 1.37× faster |
| | load | 16.8 ms | 9.8 ms | 1.72× faster |
| **hspy** 128×128, 318k vectors | size | 5873.4 KB | 4280.5 KB | **−27.1%** |
| | save | 57.9 ms | 116.1 ms | 2.00× slower |
| | load | 55.7 ms | 32.9 ms | **1.69× faster** |
| **hspy** 64×64, 101k vectors | size | 1840.2 KB | 1378.0 KB | −25.1% |
| | save | 15.1 ms | 37.1 ms | 2.47× slower |
| | load | 17.2 ms | 12.9 ms | 1.34× faster |

**On the hspy save regression** — it is compression cost, not encoding
cost. Isolating it on the 128×128 case:

| hspy 128×128 | main | this PR |
|---|---|---|
| save, `compression=None` | 53.4 ms | **16.5 ms** (3.2× faster) |
| save, `compression="gzip"` (default) | 56.7 ms | 117.1 ms |
| what gzip actually achieves | 6297 → 5873 KB (6.7%) | 6012 → 4280 KB (**29%**) |

Uncompressed, the CSR write path is 3.2× faster. With gzip it is slower
because the dense buffer gives gzip real cross-cell redundancy to exploit
— it removes 29% versus 6.7%, i.e. roughly 4× more compression work for a
27% smaller file. zspy shows the same effect without the penalty because
its default Blosc/zstd is a much faster codec, which is why zspy wins on
size *and* both directions of throughput.

So on hspy this is a genuine time/space trade, not a free win: a smaller
file for a slower save. Worth a maintainer opinion on whether that default
is the right one — it is easy to keep the current behaviour for hspy if
not.

### Forward-compatibility with zarr 3

This is the part I think matters most beyond the throughput numbers.
Because `values` and `offsets` are plain concrete-dtype arrays, CSR-encoded
data needs none of the object-codec machinery that blocks zarr 3 support
(see #352). Files written **here under zarr 2.18.7**, then opened under
**zarr 3.2.1**:

```
--- CSR encoding (this PR) ---
  OK  dtype=float64 shape=(1764, 2) sum=224525.2029
  offsets OK: shape=(257,) last=1764 (== rows: True)
  attrs: {'_ragged_encoding': 'csr', '_ragged_nav_shape': [16, 16]}

--- generic object-dtype (main) ---
  FAILED  ValueError: No Zarr data type found that matches
          {'name': '|O', 'object_codec_id': 'vlen-array'}
```

The ragged array reconstructs completely under zarr 3 with nothing but
numpy — no rsciio v3 port required for the data itself:

```python
values  = grp["data"][:]
offsets = grp["_ragged_offsets_data"][:]
nav     = tuple(grp["data"].attrs["_ragged_nav_shape"])
cells   = np.split(values, offsets[1:-1], axis=0)
# -> reconstructed ragged array (16, 16), 1764 vectors, cells intact
```

That is the practical significance: vector/marker data saved in this
encoding stays readable across the zarr 2 → 3 transition, whereas the same
data in the current encoding does not. It doesn't make rsciio zarr-3 ready
on its own — the writer still uses v2-only APIs — but it removes ragged
vector data from the set of things that transition has to solve.

### Progress of the PR
- [x] Change implemented
- [ ] update docstring (if appropriate)
- [ ] update user guide (if appropriate)
- [x] add a changelog entry in the `upcoming_changes` folder
- [ ] Check formatting of the changelog entry in the readthedocs build
- [x] add tests
- [ ] ready for review

### Minimal example of the new feature
```python
import numpy as np
import hyperspy.api as hs

data = np.empty((10, 10), dtype=object)
for index in np.ndindex(data.shape):
    n = np.random.randint(0, 8)
    data[index] = np.random.random((n, 2))  # e.g. per-position vectors

s = hs.signals.BaseSignal(data, ragged=True)
s.save("vectors.zspy")  # written via the new dense CSR encoding

s2 = hs.load("vectors.zspy", lazy=True)  # reconstructed lazily
```

### DIFF
--- a/rsciio/_hierarchical.py
+++ b/rsciio/_hierarchical.py
@@ -1010,8 +1010,13 @@ class HierarchicalWriter:
             else:
                 new_data, shapes = flatten_data(data, is_hdf5=cls._is_hdf5)
 
-            got_data = False
-            while not got_data:
+            # If the shape or dtype/etc of an existing dataset don't match,
+            # delete it and create a new one -- needed when the key previously
+            # held a CSR-encoded (dense) array, whose shape never matches the
+            # ragged one written here. Retry exactly once: a TypeError that
+            # survives the delete comes from the data itself, not from a
+            # conflicting dataset, and must propagate rather than spin.
+            for last_attempt in (False, True):
                 try:
                     dset = cls._get_object_dset(group, new_data, key, chunks, **kwds)
                     shape_dset = cls._get_object_dset(
@@ -1022,13 +1027,10 @@ class HierarchicalWriter:
                         dtype=int,
                         **kwds,
                     )
-                    got_data = True
+                    break
                 except TypeError:
-                    # Same as the concrete-dtype branch above: if the shape or
-                    # dtype/etc of an existing dataset don't match, delete it
-                    # and create a new one in the next loop run. Needed when
-                    # the key previously held a CSR-encoded (dense) array,
-                    # whose shape never matches the ragged one written here.
+                    if last_attempt:
+                        raise
                     for key_ in (key, f"_ragged_shapes_{key}"):
                         if key_ in group:
                             del group[key_]
--- a/rsciio/_hierarchical.py
+++ b/rsciio/_hierarchical.py
@@ -29,7 +29,11 @@ from rsciio._docstrings import SHOW_PROGRESSBAR_DOC
 from rsciio.utils._array import is_dask_array
 from rsciio.utils._tools import ensure_unicode
 
-version = "3.3"
+# 3.4 adds the CSR (dense values + offsets) encoding for ragged arrays with a
+# uniform numeric per-position shape. Readers older than 3.4 don't know about
+# the `_ragged_encoding` attribute and would silently misread such a dataset as
+# a plain dense array, so the bump is what makes them warn.
+version = "3.4"
 
 default_version = Version(version)
 
@@ -843,6 +847,29 @@ class HierarchicalWriter:
     def _store_data(*arg):  # pragma: no cover
         raise NotImplementedError("This method must be implemented by subclasses.")
 
+    @staticmethod
+    def _clear_ragged_state(group, key):
+        """
+        Drop any ragged-array bookkeeping left over from a previous write of
+        ``key``: the CSR marker attributes and both flavours of sidecar
+        dataset.
+
+        Rewriting a key doesn't necessarily recreate its dataset -- when the
+        new shape and dtype happen to match, ``require_dataset`` reuses the
+        existing one -- so without this the old encoding's metadata can
+        survive and make the reader reconstruct the new data the wrong way
+        (e.g. a plain dense array read back as a ragged one). Sidecars from
+        the encoding *not* used by this write would also linger as dead
+        datasets in the file.
+        """
+        if key in group:
+            for attr in ("_ragged_encoding", "_ragged_nav_shape"):
+                if attr in group[key].attrs:
+                    del group[key].attrs[attr]
+        for sidecar in (f"_ragged_offsets_{key}", f"_ragged_shapes_{key}"):
+            if sidecar in group:
+                del group[sidecar]
+
     @classmethod
     def _write_ragged_csr(
         cls, group, data, key, trailing_shape, dtype, show_progressbar=True, **kwds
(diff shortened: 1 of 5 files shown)


CSSFrancis:
@ericpre when you have a chance this should allow us to support zarr 3.0.0. It also is much faster for storing large arrays of ragged arrays.

ericpre:
@CSSFrancis, would it be possible to have a summary of the alternatives to confirm that using CSR is a suitable approach? It   sounds very sensible and it is more in the interest of having documented in this PR. Would it be worth doing an hyperspy-proposals, as this changes the format?

Would it be worth dropping support for the more generic "non-uniform" ragged array?

PR #546, opened 2026-09-11
Preserve image dtype when reading Velox EMD files

### LINKED ISSUE #476: EMD reader promotes integer data to float64
Observed on a Velox EMD file where images were stored as chunked 16bit unsigned int datasets, i.e. "u2" rosettasciio returns these under the "data" entry in the parsing results dictionary as a numpy array whose values are promoted to np.float64. While this is not necessarily in this particular case a loss in precision but in practice resulting in an eventually unnecessary quadrupling of the memory consumption.

Is there are chance to have the individual entries under the "data" entry read and loaded such that these are represented with the same datatype as in the source or is there some normalization required for reasons and context that I am missing, that demands a promotion to np.float64 ?

hueseyincelik:
### Description of the change

Fixes #476.

When a Velox EMD file is read using:

```python
from rsciio import emd

objects = emd.file_reader("image.emd")
data = objects[0]["data"]
```

integer image datasets are currently promoted to `float64` during eager loading. For a `uint16` image, this unnecessarily increases memory usage by a factor of four.

This happens because the intermediate NumPy array is allocated without a dtype. This PR allocates it using the dtype of the source HDF5 dataset, preserving the original representation. The tests now verify this behavior using the existing `uint16` Velox fixtures.

### Progress of the PR

- [x] Preserve the source dtype when reading Velox EMD images eagerly
- [x] update docstring — not required
- [x] update user guide — not required
- [x] add a changelog entry
- [ ] Check formatting of the changelog entry in the documentation build
- [x] add tests
- [x] ready for review

### Minimal example of the bug fix

```python
from rsciio import emd

objects = emd.file_reader("image.emd")
print(objects[0]["data"].dtype)  # uint16, matching the source dataset
```

### Validation

- Pre-commit checks pass.
- The eager reader preserves both the values and `uint16` dtype of the existing Velox image fixture.
- The new dtype assertions and 37 other Velox EMD tests pass. Two pre-existing unit-string checks fail locally with HyperSpy 2.4 because it normalizes `µm` to `μm`.


hueseyincelik:
Small follow-up on the old comment saying that specifying `dtype` is substantially slower: I benchmarked this locally and see the opposite for the Velox image data I tested.

Using `np.empty(h5data.shape, dtype=h5data.dtype)` avoids the implicit conversion to NumPy's default `float64` destination during `read_direct()`. For `uint16` data, that also means 4× less destination memory, so the dtype-preserving path is both faster in these benchmarks and uses substantially less memory.

With h5py 3.16.0, HDF5 2.0.0, and NumPy 2.5.3, matching-dtype `read_direct()` took about 0.014 ms per read for the existing 512 × 512 Velox fixture, compared with 0.162 ms for a default-`float64` destination. On synthetic 1024 × 1024 chunked `uint16` datasets, it took about 0.05 ms compared with 0.7 ms. These are cached microbenchmarks rather than representative end-to-end timings, but the conversion cost is consistent across the cases tested.

The old performance observation may have applied to older h5py/HDF5 versions or a different dtype or workload, but it does not seem to hold for the current case.

It may also be worth revisiting the `read_direct()` approach separately and benchmarking it against simply using `h5data[:]` with current h5py/HDF5 versions. In these tests, slicing performed similarly to matching-dtype `read_direct()`. That would be a broader change, though, and is outside the scope of this PR, which is only intended to fix dtype preservation.

